PR #7, opened 2023-04-27
Fix and finish HNN implementation

michakraus:


benedict-96 (commented):
The naming of functions in `pendulum.jl` is probably not the smartest choice. 
`get_data_set` -> `get_vector_field_data`
`pendulum_data` -> `get_phase_space_data` would probably be better. 
Maybe we could also define structs for these two functions that generate data for arbitrary systems, by e.g. providing key words or a Hamiltonian.

michakraus:
The naming is legacy and can certainly be improved upon. For the time beging we can go with
`get_data_set` -> `vector_field_data`
`pendulum_data` -> `phase_space_data`
Feel free to provide a PR.

More elaborate features and a general interface should be implemented in ReducedComplexityModeling.jl.

michakraus:
Btw, the existence of the two functions itself is pretty bad. A HNN should take a pair `(q,p)` as inputs, not `x`.

PR #35, opened 2023-07-25
Merge applica abstract nn

Theozeud:
I adapted the use of the new Optimizer Structure with the new AbstractNeuralNetwork structure. 
Now Optimizer deals with Tuple of NamedTuple instead of NamedTuple of NamedTuple.
I commented on some tests of Benedikt because they are not adapted anymore with the new AbstractNeuralNetwork.
I also adapted the gradient Layer in the framework of the dense layer of AbstractNeuralNetwork.jl in the initialization process and to get rid of Lux.
I also adapted the training process with the new AbstractNeuralNetwork structure.
There is also an example of utilization of tools of this package on harmonic oscillator with GSympNet.
The HNN doesn't work anymore due to zygote issues but should be fixed soon.


benedict-96:
Got rid of most of the Lux dependencies, a few remain but are easy to fix (also wrote some documentation in the process). 

Small comment about the Gradient layer: 
In `AbstractNeuralNetworks` the layer type stores the input and output dimension. The Gradient layer also has two dimensions, but the second one is an internal "rescaling dimension" that is independent of the in- and out-put dimension of the network. I have therefore changed the layer struct to: 
```
struct Gradient{M, N, full_grad, change_q, AT} <: AbstractExplicitLayer{M, N}
        second_dim::Integer
        activation::AT
end
```

michakraus (approved):

PR #50, opened 2023-08-21
Transformer utils

benedict-96:
This PR includes the new activation function for structure-preserving transformers (`orthonormal_activation_cayley`) and also some utils that make it easier to handle data in tensor form (folder `data_loader`; we should probably give this a different name).


michakraus:
Please remove the images from the commit and in general make sure not to check in binary files (images, hdf5, jld2).

benedict-96:
Removed binary files and `orthonormal_activation_old`.

PR #51, opened 2023-08-22
Clean up

benedict-96:
Removed two zombie files and wrote some documentation.


benedict-96:
Removed `png`s and `jdl2`.

PR #56, opened 2023-08-26
Simplify retractions

benedict-96:
I simplified retractions for manifold layers (if a layer includes manifolds, it is now a sub type of `LayerWithManifold` or `LayerWithOptionalManifold`). I think this is branching off from `attentionlayer_tests` instead of `main`. Sorry about that. Won't happen again.


michakraus:
Please rebase to main.

PR #66, opened 2023-09-01
Add diagram to doc

benedict-96:
Added new documentation and cleaned up the old one. I also increased the tolerance for one test which lead to a failed CI run in the previous PR.


benedict-96:
Pushed again into this branch (I just added some more documentation). This PR only affects `docs`.

PR #71, opened 2023-09-07
Mnist script

benedict-96:
Added Mnist utils in order to process it with `DataLoader`. Also includes some tests.


benedict-96:
Pushed again into that branch. There was an issue with GPU compatibility. It should be fixed now.

PR #84, opened 2023-10-20
Batch struct

benedict-96:
Introduced a new struct `Batch` that takes care of drawing new batches for a new epoch during training. This was previously (sloppily) done with `DataLoader`. Some helper functions have also been added, most importantly `optimize_for_one_epoch(::opt::Optimizer, model, ps, dl::DataLoader, batch::Batch, loss)`. I also tried to add and improve documentation for routines I changed/introduced.


michakraus:
There is some Julia v1.8 issue that needs to be fixed, plus a rebase.

benedict-96:
Tests for `v1.8` go through now. I replaced the outer by an inner constructor and now it works: 

```julia
struct Batch{seq_length}
    batch_size::Integer
    seq_length::Union{Nothing, Integer}
end

function Batch(batch_size, seq_length)
    Batch{true}(batch_size, seq_length)
end

function Batch(batch_size::Integer)
    Batch{false}(batch_size, nothing)
end
end
```

to

```julia
struct Batch{seq_length}
    batch_size::Integer
    seq_length::Union{Nothing, Integer}

    function Batch(batch_size, seq_length)
        new{true}(batch_size, seq_length)
    end

    function Batch(batch_size::Integer)
        new{false}(batch_size, nothing)
    end
end
```

michakraus:
I see two issues here:
1) You use `seq_length` as type parameter and as a field name.
2) Your fields do not have concrete types.

I would rather write this along the following lines:
```
struct Batch{seq_type <: Union{Nothing, Integer}}
    batch_size::Int
    seq_length::seq_type

    function Batch(batch_size, seq_length = nothing)
        new{typeof(seq_length)}(batch_size, seq_length)
    end
end

hasseqlength(::Batch{<:Integer}) = true
hasseqlength(::Batch{<:Nothing}) = false
```

benedict-96:
Copy-pasted the suggested change and pushed again.

michakraus:
I guess you were a bit too quick and forgot to run your tests 😉 .

There are a few more changes required, e.g.,
```
function (batch::Batch{false})
```
needs to become
```
function (batch::Batch{Nothing})
```
etc.

benedict-96:
sorry :sweat_smile: . Should be working now.

PR #98, opened 2023-12-13
Multiplication and addition tests for custom arrays

benedict-96:
Added tests for addition, scalar multiplication and matrix multiplication of custom arrays. Tried to make tests in `runtests.jl` more readable.


michakraus:
Unfortunately, we need a rebase here.

PR #109, opened 2024-01-08
Fix version in the documenter to 1 and fixed cpu allocation problem.

benedict-96:
I fixed the julia version to 1.9, this is the same version I have locally. Apart from this I don't see any differences (locally the html files are rendered correctly).


michakraus:
It would be better to set this to `version: '1'` so that the latest release is used at all times without the need for future updates.

Other than that, there are a few changes in here which do not really match the title of the PR 😉.

PR #121, opened 2024-03-05
Remove unnecessary dependencies

benedict-96:
Removed (i) CUDA (ii) ForwardDiff (iii) GPUArrays (iv) Lux. Regarding (iv) from `Project.toml`. The CUDA dependency meant that the tests for nightly didn't go through; the hope was that they are going through now (but it seems Zygote is posing a similar problem). I've also commented out some of the functions in `backends/lux.jl` to be able to remove (iv). I'm not sure if we should keep this backend and, if we want to keep it, if we have to import Lux to do so.


benedict-96:
The documentation workflow is also not going through for now, but this is because the interface in `BrenierTwoFluid` has changed. This is taken care of by https://github.com/JuliaGNI/GeometricMachineLearning.jl/pull/120

benedict-96:
Just resolved the merge conflict that appeared as a result of #120 ...  the CI tests (except nightly probably, but including the Documenter) should go through now.

PR #131, opened 2024-04-12
Transformer integrator

benedict-96:
Introduced a new type `TransformerIntegrator` that makes the transformer behave similarly to a classical multi-step method. For now this is done by extending `iterate`, but will have to be interfaced with `GeometricIntegratorsBase` eventually.


src/architectures/regular_transformer_integrator.jl
michakraus, line 1: There's a typo: "the defualt is"

benedict-96:
Fixed the typo and adjusted the interface in the docs which meant the Documenter workflow was crashing before.

PR #138, opened 2024-04-24
Symplectic Autoencoder and PSD architecture

benedict-96:
Implemented a symplectic autoencoder and a PSD architecture. For the PSD architecture we don't perform training but use `solve!`. They are both subtypes of `AutoEncoder`. Much of the functionnality for `SymplecticAutoencoder` and `PSDArch` is performed by this generic `AutoEncoder` type.


michakraus:
Please rename `get_encoder`, `get_decoder` and similar functions and remove the `get_`.

benedict-96:
I thought that the `get_` may be appropriate in this particular case because we don't actually compute anything or allocate new memory. The `SymplecticAutoencoder` stores the parameters for the encoder and the decoder (they are trained simultaneously) and after training we can "get" the encoder (or decoder) by using `get_encoder` (or `get_decoder`). We could however also rename `get_encoder` -> `obtain_encoder` for example.

michakraus (approved):

benedict-96:
Changed names of functions:
* `get_encoder` -> `encoder`
* `get_decoder` -> `decoder`

PR #181, opened 2024-12-11
Removed geometric integrators from dependencies.

benedict-96:


benedict-96:
I opened two new issues concerning re-implementation of stuff that is removed in this pr:
1. https://github.com/JuliaGNI/GeometricMachineLearning.jl/issues/182
2. https://github.com/JuliaGNI/GeometricMachineLearning.jl/issues/183

PR #207, opened 2025-07-15
Parametric Generalized Hamiltonian Neural Networks

benedict-96:
Implements parametric generalized Hamiltonian neural networks (PGHNNs): architectures whose forward pass takes the parameters of the *system* alongside the state, so one network covers a parameter range rather than a single problem instance.

38 files, +2322 / −152 against `main`.

## What it adds

`GeneralizedHamiltonianArchitecture` is implemented — it used to be a stub whose constructor threw `error("GHNN still has to be implemented!")`. It composes `n_integrators` symplectic Euler steps, each differentiating a learned kinetic or potential energy. Around it: `ForcedGeneralizedHamiltonianArchitecture` and `ForcedSympNet` with `ForcingLayer`s, `ParametricDataLoader`, `ParametricLoss`, a `SymbolicPullback` for the parametric case, `ParametricResNet` as the non-structure-preserving baseline, and `QPT2`/`QPTOAT2`.

## Brought up to today's `main`

The branch was three weeks and 160 commits behind. `main` is merged in rather than rebased onto, so the four commits of actual work stay readable. Eleven files conflicted; `main`'s version is taken in each and the branch's deltas re-applied, which is what keeps `main`'s whole-tree reformat out of this diff.

**The breaking change is gone.** `main` retired the `train!` subsystem, so `src/training_method/`, `test/train!/test_method.jl` and `test/training_phnn.jl` go with it. The rename of `SymplecticEuler` to `SymplecticEulerIntegrator` existed only to free that name for the new layer type, and `main` no longer defines it at all. `Distances` drops out too: `sqeuclidean` served only those training methods.

**`Project.toml` is byte-identical to `main`'s** — AbstractNeuralNetworks 0.8, SymbolicNeuralNetworks 0.8, NeuralNetworkParameters 0.3, GeometricOptimizers 0.7, Zygote 0.7, julia 1.11. `[extras]`/`[targets]` moved to `test/Project.toml` on `main`, and the branch follows.

Two adaptations beyond the merge itself. `_eltype` is gone from `main`, replaced by `NeuralNetworkParameters.parameter_eltype`; the one call site that needed a per-trajectory check keeps a small local helper. And the five new test files are wired into `main`'s per-subject drivers rather than a flat `runtests.jl`, which `test/reachability.jl` requires.

## Seven bugs, all on paths nothing ran

`test/training_phnn.jl` was never in `runtests.jl`, so the loop this PR is *for* had never executed, and six of the new files were at 0 % coverage. Writing a construct-and-evaluate test for each turned up:

1. `concatenate_array_with_parameters(::AbstractMatrix, ::AbstractVector)` used `vcat` where it needs `hcat`, collapsing a batch into one long vector.
2. `optimize_for_one_epoch!` called `_unpack_tuple`, which has never existed in this package.
3. `Zygote` differentiates *through* the `NeuralNetworkParameters` struct, so every nesting level of the gradient comes back wrapped in `(params = …,)`. These architectures nest — a `SymplecticEuler` layer holds a whole sub-network — and `_get_params` only unwraps the top, so `_unwrap_gradient` recurses.
4. **`ForcedGeneralizedHamiltonianArchitecture` could not be evaluated at all.** The parameter-dependent `NeuralNetwork` functor and the `Optimizer` entry point were written for `GeneralizedHamiltonianArchitecture`, and the forced architecture is a *sibling* of it under `HamiltonianArchitecture`, not a subtype — so `nn(x, μ)` fell through to AbstractNeuralNetworks' generic functor, which read the system parameters as the network parameters and reached the first layer as a `Float64`.
5. `ParametricResNet(::DataLoader, n_blocks, width; parameters = …)` accepted `parameters` and did not forward it, always building a network with no parameter dependence.

Two more surfaced while making the scripts run, and are new in this catch-up:

6. **`NeuralNetwork{<:ParametricResNet}` had no parameter-dependent functor either** — the same defect as (4), one level along. A `ParametricResNet` is a `NeuralNetworkIntegrator`, which neither existing method covers. The existing test called `apply_parametric` directly, which is why it went unseen; it now also asserts the functor agrees with it.
7. **`_processing` only unwrapped the top of the tree**, so training *any* nested architecture on an ordinary `DataLoader` raised `FieldError: type NamedTuple has no field L1` inside `_tree_optim_step!`. `_unwrap_gradient` — added for (3) but wired only into the `ParametricDataLoader` path — is the fix: `_processing` is `_unwrap_gradient ∘ _get_params ∘ _get_contents` now. Covered by the `ForcedSympNet` testsets.

## Flattening the system parameters

The system parameters are flattened into the network input. The branch used `ParameterHandling` and pirated three `flatten` methods on it. That could not work: GeometricOptimizers defines `ParameterHandling.flatten(x)` with an unbound type parameter, and *that* method wins — `UndefVarError: T not defined in static parameter matching`.

`flatten`/`unflatten` from [NeuralNetworkParameters](https://github.com/JuliaGNI/NeuralNetworkParameters.jl) replace it, and because AbstractNeuralNetworks exports `params` this needs **no** method on a foreign type; the layout a layer stores is a value rather than a closure.

## Type piracy

Six of the nine methods are gone. Three were fixed upstream and are simply deleted: `applychain` (AbstractNeuralNetworks 0.7 leaves its data argument untyped *for this reason*), `networkbackend(::LazyArrays.ApplyArray)` (LazyArrays is not a GML dependency any more) and `h5save(::HDF5.Group, ::NetworkParameters, …)`. The four three-argument `Chain` functors became `apply_parametric`, a function GML owns.

What is left is genuinely upstream's: `SymbolicNeuralNetworks.Jacobian(f, nn, dim2)`, `Jacobian(nn, dim2)`, and three `SymbolicPullback` functors whose third tuple element carries the system parameters. `test/aqua.jl`'s gate goes from 12 to 17 with all five named and attributed to #243, rather than the check being switched off.

## `scripts/`

`main` imposed a three-directory layout with a CI gate. The three training scripts move to `scripts/reproduction/` and their shared helper to `scripts/utilities/`, so `Scripts.yml` runs them in smoke mode; they carry `smoke_size` constants like every other reproduction script. They arrived as top-level files that nothing ran, and `Train_DampedOscillator_QP.jl` did not parse at all — an unbalanced `end`. A one-off probe that printed two parameter counts is dropped.

Running them is what uncovered bugs 6 and 7.

## On #245

SymbolicNeuralNetworks 0.6 shipped the layerwise pullback, which its changelog says fixes this case. Measured, it does not reach it: SNN's seam was a plain `Vector{Num}`, so a `SymplecticEuler` built with `return_parameters = true` cannot be seeded. The `n_integrators > 1` guard therefore stays here, and #251 — stacked on this branch — is what lifts it.

## Verification

`Pkg.test()` exits 0 on this branch, with Aqua's block at 10/10.

| testset | assertions |
|---|---|
| `Generalized Hamiltonian Neural Network` | 2 |
| `Symbolic pullback for a single-layer PGHNN` | 10 |
| `PGHNN training on a ParametricDataLoader` | 4 |
| `Parametric and forced layers and architectures` | 68 |
| `Parametric DataLoader` | 2 |

All three reproduction scripts run end to end under `GML_SMOKE`.

The pre-catch-up tip of this branch was `6db0b980`, and the tip before that `43d89206`; both stay reachable by SHA.


scripts/TimeDependentHarmonicOscillatorForcedGeneralizedHamiltonianNeuralNetwork.jl
```diff
@@ -0,0 +1,161 @@
...
+# Generating the solution array
+ni = ni_dim^2
+q  = zeros(Float64, ni, nt+1)
+p  = zeros(Float64, ni, nt+1)
+t  = collect(dt * range(0, nt, step=1))
+
+"""
+Turn a vector of numbers into a vector of `NamedTuple`s to be used by `ParametricDataLoader`.
+"""
+function turn_parameters_into_correct_format(t::AbstractVector, IC::AbstractVector{<:NamedTuple})
```
benedict-96: This functionality should most likely be in `GeometricProblems` or `GMLDataSets`.

scripts/TimeDependentHarmonicOscillatorForcedGeneralizedHamiltonianNeuralNetwork.jl
```diff
@@ -0,0 +1,161 @@
...
+	vec_of_params = NamedTuple[]
+	for time_step ∈ t
+		time_step == t[end] || push!(vec_of_params, (t = time_step, ))
+	end
+	vcat((vec_of_params for _ in axes(IC, 1))...)
+end
+
+for i in 1:nt+1
```
benedict-96: The data set should be generated with `GeometricProblems` (I would think). The problem may be that there is no good way at the moment to obtain analytic solutions (which is what happens here).

scripts/TimeDependentHarmonicOscillatorForcedGeneralizedHamiltonianNeuralNetwork.jl
```diff
@@ -0,0 +1,161 @@
...
+p = [1.5 2.5 3.5; 4.5 5.5 6.5]
+qp = (q = q, p = p)
+turn_q_p_data_into_correct_format(qp)
+
+# output
+
+(q = [1.0 2.0; 4.0 5.0;;; 2.0 3.0; 5.0 6.0], p = [1.5 2.5; 4.5 5.5;;; 2.5 3.5; 5.5 6.5])
+```
+"""
+function turn_q_p_data_into_correct_format(qp::QPT2{T, 2}) where {T}
```
benedict-96: This should be in `GMLDataSets` (I think).

scripts/TimeDependentHarmonicOscillatorForcedGeneralizedHamiltonianNeuralNetwork.jl
```diff
@@ -0,0 +1,161 @@
...
+# attrs(h5)["ni"] = ni
+# attrs(h5)["nt"] = nt
+# attrs(h5)["dt"] = dt
+# 
+# close(h5)
+
+"""
+This takes time as a single additional parameter (third axis).
+"""
+function load_time_dependent_harmonic_oscillator_with_parametric_data_loader(qp::QPT{T}, t::AbstractVector{T}, IC::AbstractVector) where {T}
```
benedict-96: Again, probably better in `GMLDataSets`.

scripts/TimeDependentHarmonicOscillatorParametricResnet.jl
```diff
@@ -0,0 +1,150 @@
...
+# Generating the solution array
+ni = ni_dim^2
+q  = zeros(Float64, ni, nt+1)
+p  = zeros(Float64, ni, nt+1)
+t  = collect(dt * range(0, nt, step=1))
+
+"""
+Turn a vector of numbers into a vector of `NamedTuple`s to be used by `ParametricDataLoader`.
+"""
+function turn_parameters_into_correct_format(t::AbstractVector, IC::AbstractVector{<:NamedTuple})
```
benedict-96: This seems to be a copy of the code that I commented already above. We should try to avoid duplicates like this (and probably outsource this functionality to `GMLDataSets`).

scripts/TimeDependentHarmonicOscillatorParametricResnet.jl
```diff
@@ -0,0 +1,150 @@
...
+# attrs(h5)["ni"] = ni
+# attrs(h5)["nt"] = nt
+# attrs(h5)["dt"] = dt
+# 
+# close(h5)
+
+"""
+This takes time as a single additional parameter (third axis).
+"""
+function load_time_dependent_harmonic_oscillator_with_parametric_data_loader(qp::QPT{T}, t::AbstractVector{T}, IC::AbstractVector) where {T}
```
benedict-96: Functionality should be in `GMLDataSets`. Confer comment above.

scripts/Train_DampedOscillator_QP.jl
```diff
@@ -0,0 +1,114 @@
...
+T          = 13
+nt         = 100                    # number of time steps
+dt         = T/nt                    # time step
+n_epochs   = 100000
+n_epochs   = 3
+width      = 4                      # width of the neural network
+nhidden    = 3                       # number of hidden layers in the neural network
+batch_size = 5000                    # the size of the batch
+
+path_out = "D:\\RESEARCH - UTWENTE\\GFHNNs\\Damped Oscillator\\network_TEST.jld2"
```
benedict-96: We should proably not have the local directory in there. We should delete this wherever it occurs ...

scripts/Train_DampedOscillator_QP.jl
```diff
@@ -0,0 +1,114 @@
...
+#path_out = "/home/tyranowskitm/GFHNNs/DampedOscillator/OUTPUTS/network.jld2"
```
benedict-96: confer comment above.

scripts/Train_DampedOscillator_QP.jl
```diff
@@ -0,0 +1,114 @@
...
+
+# Generating the solution array
+ni    = ni_dim^2
+omega = sqrt(4-nu^2) / 2
+
+q  = zeros(Float64, ni, nt+1)
+p  = zeros(Float64, ni, nt+1)
+t  = collect(dt*range(0,nt,step=1))
+
+for i in 1:nt+1
```
benedict-96: Analytic solution here seems to be a duplicate of code that appeared above. Confer that bit for relevant comments.

src/data_loader/parametric_data_loader.jl
```diff
@@ -0,0 +1,143 @@
...
+        for dim in 1:sys_dim 
+            data.q[dim, :, i] = solution.q[:, dim]
+            data.p[dim, :, i] = solution.p[:, dim]
+        end 
+    end
+
+    ParametricDataLoader(data, params)
+end
+
+# """
```
benedict-96: Is this still needed (potentially). If it is obsolete we can just delete it.

src/data_loader/parametric_data_loader.jl
```diff
@@ -0,0 +1,143 @@
...
+#     rearrange_parameters(parameters)
+# 
+# Rearrange `parameters` such that they can be used by [`ParametricDataLoader`](@ref).
+# """
+# function rearrange_parameters(parameters::Vector{<:NamedTuple})
+#     parameters_rearranged = zeros(_eltype(parameters), )
+# end
+
+# function batch_over_two_axes(batch::Batch, number_columns::Int, third_dim::Int, dl::ParametricDataLoader)
```
benedict-96: Confer comment above.

src/layers/parametric_resnet_layer.jl
```diff
@@ -0,0 +1,64 @@
...
+    input = concatenate_array_with_parameters(x, problem_params)
+    x + d.activation.(ps.downscale_weight * d.activation.(ps.upscale_weight * input .+ ps.upscale_bias) .+ ps.bias)
+end
+
+function (d::ParametricResNetLayer{M, M, F, PT, true})(x::AbstractVecOrMat, problem_params::OptionalParameters, ps::NamedTuple) where {M, F, PT}
+    input = concatenate_array_with_parameters(x, problem_params)
+    (x + d.activation.(ps.downscale_weight * d.activation.(ps.upscale_weight * input .+ ps.upscale_bias) .+ ps.bias), problem_params)
+end
+
+# function (d::ParametricResNetLayer{M, M, F, PT, false})(x::AbstractArray{T, 3}, problem_params::AbstractVector, ps::NamedTuple) where {M, F, PT, T}
```
benedict-96: Is this obsolete? Confer comment above.

src/pullbacks/symbolic_hnn_pullback.jl
```diff
@@ -25,3 +25,72 @@ function SymbolicPullback(arch::HamiltonianArchitecture)
...
+This is `SymbolicNeuralNetworks.SymbolicPullback(nn, loss)` with the system parameters threaded
+through. `build_nn_function` generates a function of *one* input array, so the flattened system
+parameters are appended to the network input, and the symbolic expression splits them off again with
+[`_flatten_system_parameters`](@ref) and `unflatten`. The numeric side does the same concatenation,
+in the call operators below.
+
+`reduce = +`: the loss of a batch is the sum of the losses of its samples, so its gradient is the
+sum of the per-sample gradients.
+"""
+function SymbolicPullback(nn::NeuralNetwork, loss::ParametricLoss,
```
benedict-96, line 94: With the `SymbolicPullback` there seem to be massive performance issues that have to be investigated. Also: should parts of this be in `SymplecticNeuralNetworks`?

src/utils.jl
```diff
@@ -165,9 +174,82 @@ This could be data in ``(q, p)\in\mathbb{R}^{2d}`` form or come from an arbitrar
...
+`NeuralNetworkParameters.ParameterLayout` that puts such a vector back into the original shape.
+
+The parameter-dependent architectures — [`GeneralizedHamiltonianArchitecture`](@ref) and the layers
+it is built from — feed the system parameters to the network as extra input components, so they have
+to be a vector. `NullParameters` flattens to an empty one, which makes the parameter-free case fall
+out of the same code path.
+
+The layout is a *value*, not a closure, so a layer can store it in a field and stay inferable.
+"""
+_flatten_system_parameters(parameters::NamedTuple) = flatten(parameters)
```
benedict-96, line 204: This is used for symbolic expressions. It is probably debatable if this is the cleanest way of implementing this functionality. May deserve a new issue but may not be solved in this pr.

benedict-96 (commented):
Commented code with the thought of handing this review over to `claude` in the first iteration.
Something that I haven't mentioned yet is missing documentation on the parametric forced generalized Hamiltonian neural network, the forced SympNet and the WideResNet. For making this documentation more visually appealing I could provide the tikz pictures that are part of the paper pertaining to this pr.

michakraus:
## Force-pushed a rebase onto `main` (0.5.0)

The previous tip was `43d89206`; nothing is lost, that commit is still reachable by SHA on GitHub and I have it as a local branch. The branch is now one commit, `633878f`, and the PR is `MERGEABLE` again.

### What was dropped, and why

| commit | verdict |
|---|---|
| `cc1a786f` "Fix GML compilation and add generated artifacts" | **dropped.** 123 MB of build output — `docs/build.zip` (77 MB), `docs/build 2.zip` (47 MB), eleven LaTeX `.aux`/`.log`/`.fls`/`.fdb_latexmk` files, two `go_migration_inspection_*.txt` — and about four lines of source. `.gitignore` already covers all of it bar the inspection logs. The four lines are kept. It did not, in fact, fix compilation. |
| `8d243592` "Resolve merge conflicts" | **dropped.** Not this branch's work: it is `main`'s own HDF5 migration re-applied by hand. `ext/HDF5Ext.jl`, `test/hdf5_support.jl`, `docs/utils/adjust_image_size.jl`, `scripts/convert_jld2_to_h5.jl`, `src/architectures/symplectic_autoencoder.jl` and `src/layers/bias_layer.jl` were byte-identical blobs to `main`'s. Redundant once rebased. |
| `43d89206` "Merge main into…" | **dropped.** It merged `main` only as far as `d07b4c26`, so the branch never saw the GeometricOptimizers separation, and its resolution left `src/GeometricMachineLearning.jl` half from each side: `go_bridges.jl`, `src/arrays/` and `src/manifolds/` `include`d again next to `main`'s `import GeometricOptimizers`. That is the `Method overwriting is not permitted during Module precompilation` that all 13 jobs were failing on. |
| `4c0e2684` "Fix Hamiltonian symbolic code generation" | **kept.** A real seven-line fix — the `Jacobian` broadcast (`dx.(scalarize(f))` rather than `scalarize(dx(f))`) and `inplace = false` on two `build_nn_function` calls. |

The other 53 commits are collapsed into one, still authored by @benedict-96.

**Why the merges kept going wrong:** `main` reformatted `src/GeometricMachineLearning.jl` from a 4-space-indented module body to column 0, and this branch never did — so *every* merge of `main` conflicts on the whole file, and each has been resolved a bit differently by hand. Here I took `main`'s version as-is and re-applied only this branch's `include`s and `export`s. Worth keeping in mind for any other long-lived branch.

Nothing that moved to GeometricOptimizers came back: `git diff --name-only main...HEAD -- src/arrays src/manifolds src/optimizers/go_bridges.jl test/manifolds test/optimizers` is empty, and no file `main` deleted is present.

### ⚠️ This cannot merge until NeuralNetworkParameters.jl is registered

Every job now fails with one error, at dependency resolution — nothing compiles or tests wrong:

```
ERROR: LoadError: expected package `NeuralNetworkParameters [67f4d93a]` to be registered
```

The branch used `ParameterHandling` to flatten the system parameters into the network input, and pirated three `flatten` methods on it. That does not work at all: GeometricOptimizers defines `ParameterHandling.flatten(x)` with an unbound type parameter, and *that* method is the one that wins —

```
ERROR: UndefVarError: `T` not defined in static parameter matching
  [1] flatten(x::@NamedTuple{})
    @ GeometricOptimizers …/src/optimizers/named_tuple_wrapper.jl:14
```

which is D6 in [NeuralNetworkParameters' `PLAN.md`](https://github.com/JuliaGNI/NeuralNetworkParameters.jl/blob/main/PLAN.md), hit in practice. `flatten`/`unflatten` from NeuralNetworkParameters replace it — and because AbstractNeuralNetworks already exports `params`, that needs **no** method on a foreign type, so three pirated methods disappear rather than move. GML supports `julia = "1.10"`, where `[sources]` does not exist, so registering NNP 0.1.0 is a hard prerequisite.

### Three bugs, all on the training path

`test/training_phnn.jl` is not in `runtests.jl`, so the loop this PR exists for had never run:

1. `concatenate_array_with_parameters(::AbstractMatrix, ::AbstractVector)` used `vcat` where it needs `hcat`, collapsing a batch into one long vector.
2. `optimize_for_one_epoch!` called `_unpack_tuple`, which has never existed in this package.
3. `Zygote` differentiates *through* the `NeuralNetworkParameters` struct, so every nesting level of the gradient comes back wrapped in `(params = …,)`. These architectures nest — a `SymplecticEuler` layer holds a whole sub-network — and `_get_params` only unwraps the top, so `_unwrap_gradient` recurses.

New `pghnn_training_test.jl` covers that path. The two existing tests are rewritten: `parametric_data_loader_test.jl` pinned *which* shuffled batch holds which parameters, which depends on the Julia version's RNG stream, and now asserts the correspondence itself; `pghnn_symbolic_pullback_single_layer_test.jl` had no `@test` at all and re-defined in the test file what `src/` now provides, and now checks the symbolic gradient against the summed per-sample `Zygote` gradients — they agree to 4e-16.

### Dependency drift

The branch predates SymbolicNeuralNetworks 0.5: `symbolize!` is gone (`symbolic_variables` replaces it) and symbolic variables are scalar `Num`s rather than `Symbolics.Arr`s ([SNN#14](https://github.com/JuliaGNI/SymbolicNeuralNetworks.jl/issues/14)). The parametric `SymbolicPullback` is rewritten in the shape of `SymbolicNeuralNetworks.SymbolicPullback`, and five `Symbolics.Arr` special cases — three of them piracy — became unreachable and are deleted. `GeometricProblems.default_parameters` is a function since 0.8. `src/optimizers/optimizer.jl` keeps `main`'s version, since the 0.5.0 rewrite covers what the branch added there.

### The three items from the original description

- **Type piracy** — removed where it was free (`Base.NamedTuple(::NeuralNetworkParameters)` is just `params`; the `ParameterHandling` and `Symbolics.Arr` methods went with their callers). The rest each carry a `TODO` naming their proper home and are collected in #243; they need releases of AbstractNeuralNetworks and SymbolicNeuralNetworks, so they are deliberately deferred. NNP's `PLAN.md` did not cover any of this — the two entries that are *its* to own are now D9 and D10 there.
- **`QPT2`** — introduced and documented, `QPT` widened. Replacing `QPT` with it is still open.
- **Parameter-dependent `DataLoader`** — done, as `ParametricDataLoader`.

### Verified locally

`Pkg.test()` passes (exit 0; the four new testsets are 2 + 8 + 4 + 2 assertions, no failures) and `docs/make.jl` builds clean, both against a `dev`ed NeuralNetworkParameters.

michakraus:
Thanks — worked through all of these. Summary, then the two where the answer turned out to be different from what the comment assumed.

## Done

**Dead code** (`parametric_data_loader.jl:52`, `:61`, `parametric_resnet_layer.jl:38`) — all three deleted. They were not just unused but obsolete: `rearrange_parameters` was never finished (`zeros(_eltype(parameters), )` has no dimensions) and never called; the commented `batch_over_two_axes` is superseded by the `n_batches::Integer` method in `batch.jl` that the live code calls; and the commented three-dimensional `ParametricResNetLayer` methods were *wrong* — both compute `input` and then use `x` — so they were not a starting point for a real implementation either.

**The `D:\RESEARCH - UTWENTE\…` path** — gone, now `joinpath(get(ENV, "GML_OUTPUT_DIR", @__DIR__), "damped_oscillator_network.jld2")`. It was the only absolute path in the PR.

**Script duplication** — you were right, and it was worse than the comments suggest: `turn_q_p_data_into_correct_format` was *byte-identical* in all three scripts, and the two `TimeDependentHarmonicOscillator*` scripts shared a ~110-line prelude that differed only in the constants `F` and `T`. All of it now lives in `scripts/parametric_data_helpers.jl`, including the shared analytic solution as `forced_harmonic_oscillator_solution(t, IC; omega, Omega, F)` — about 200 lines less duplication. The file's header says it is a staging post, and **GMLDatasets [#5](https://github.com/JuliaGNI/GMLDatasets.jl/issues/5)** tracks moving the helpers where they belong; I wrote up there the two open questions, including your point that the data has to come from closed-form solutions because there is no good way to get analytic solutions out of `GeometricProblems` at the moment.

**Missing documentation** — partly. `ForcedGeneralizedHamiltonianArchitecture` and `ForcedSympNet` already had docstrings and manual entries. But `WideResNetLayer`, `ParametricResNetLayer` and `ParametricResNet` had **no docstring at all** and were absent from the manual; they slipped past `missing_docs` because Documenter only checks bindings that already carry one. All three are written and in the manual now. The *narrative* documentation — and your tikz figures from the paper, which I would very much like — I have deliberately left; say the word and I will write the section around them.

## `_flatten_system_parameters` (`utils.jl:204`)

Agreed it is debatable. One thing that may change the picture: it is no longer `ParameterHandling`. That could not work at all — `GeometricOptimizers` defines `ParameterHandling.flatten(x)` with an unbound type parameter and *that* method wins, so any call errors with `UndefVarError: T not defined in static parameter matching`. It is `NeuralNetworkParameters.flatten`/`unflatten` now, which also means the layer stores a `ParameterLayout` — a value, so inferable and storable — rather than a closure. Happy to open an issue if you still want the design revisited.

## `SymbolicPullback` performance — measured, and it is the other way round

This is the one worth reading. Per call the symbolic pullback is **100–1000× faster** than Zygote, and flat:

| network (`dim`, `width`, `n_int`) | params | `ZygotePullback` | `SymbolicPullback` |
|---|---|---|---|
| 2, 2, 1 | 32 | 10.4 ms | **0.1 ms** |
| 2, 4, 1 | 80 | 23.6 ms | **0.1 ms** |
| 4, 4, 1 | 88 | 33.6 ms | **0.1 ms** |
| 4, 4, 2 | 176 | 140.1 ms | — never builds |

What explodes is the **build**, and multiplicatively in `n_integrators`. Phase by phase at `dim = 4, width = 4, nhidden = 1`:

| phase | `n_int = 1` | `n_int = 2` |
|---|---|---|
| build the symbolic loss | 0.87 s, **3.4 ⋅ 10⁵ chars** | 0.03 s, **1.4 ⋅ 10⁹ chars** |
| `symbolic_derivative` | 0.66 s, **1.5 ⋅ 10⁸ chars** | never finishes |
| `build_nn_function` | 1.34 s | — |

The loss expression grows ~4,100× going from one integrator to two, *before* any differentiation, and differentiating multiplies by a further ~440×. At two integrators the process passes 8 GB and does not return.

**Cause**: each `SymplecticEuler` layer's forward pass calls the executable gradient `build_gradient` produced for its energy network. Traced on symbolic `Num`s that inlines the entire gradient expression, so stacking integrators inlines it inside itself. `cse = true` cannot help — it runs at code generation, long after the expression has been materialised.

**Fix**: stop tracing the inlined chain and compose the pullback layer by layer from the gradients each `SymplecticEuler` has already built, which keeps the expression linear in depth. That answers your other question too — yes, parts of this belong in `SymbolicNeuralNetworks`: it needs either a layerwise pullback protocol there, or a `build_nn_function` that can take a composition without inlining it. Written up with all the numbers in **[#245](https://github.com/JuliaGNI/GeometricMachineLearning.jl/issues/245)**.

I did not fix it here — it is a redesign of the symbolic pipeline, not a tweak. What this PR does is make the failure honest: `SymbolicPullback(nn, ::ParametricLoss, μ)` now throws an `ArgumentError` for `n_integrators > 1` explaining why and pointing at `ZygotePullback`, instead of appearing to hang. One integrator builds in ≈1.4 s and is fully supported; the default training path uses `ZygotePullback` and is unaffected.

Everything is on `2463e03b`; `Pkg.test()` and `docs/check_references.jl` both pass locally.

PR #230, opened 2026-05-07
Replace GML optimizer internals with GeometricOptimizers backend

benedict-96:
## Summary

This PR removes the optimizer implementations that were duplicated inside `GeometricMachineLearning` and replaces them with the [`GeometricOptimizers`](https://github.com/JuliaGNI/GeometricOptimizers.jl) package as the authoritative backend.

### What changed

**Deleted optimizer source files (~850 lines removed):**
- `src/optimizers/adam_optimizer.jl`
- `src/optimizers/adam_optimizer_with_learning_rate_decay.jl`
- `src/optimizers/bfgs_cache.jl`, `bfgs_optimizer.jl`
- `src/optimizers/gradient_optimizer.jl`, `momentum_optimizer.jl`
- `src/optimizers/optimizer.jl`, `optimizer_caches.jl`, `optimizer_method.jl`
- `src/optimizers/manifold_related/` (global sections, retractions, retraction types, modified exponentials)

**New `src/utils.jl` integration layer (~250 lines added):**
- `GMLEuclideanState` — lightweight Adam/Momentum/Gradient state for plain-array (non-manifold) parameters, avoiding the need to wrap Euclidean tensors in GO's manifold machinery.
- `_make_optimizer_cache` / `_make_optimizer_state` — recursive dispatchers that walk the `NeuralNetworkParameters` tree, routing manifold leaves (e.g. `StiefelManifold`) to GO's `AdamCache`/`MomentumCache`/`GradientCache` and Euclidean leaves to `GMLEuclideanState`.
- `_adapt_method_to_T` — promotes the optimizer method's numeric type to match the parameter's element type, so `Adam()` (which defaults to `Float32`) works correctly with `Float64` network parameters.
- `_leaf_optim_step!` / `_tree_optim_step!` — dispatchers that call GO's `update!` for manifold caches and GML's own update rules for Euclidean caches.
- `AdamOptimizerWithDecay` is retained as a GML-specific extension of GO's `OptimizerMethod` — **superseded upstream, see "Follow-up upstream" below**.

**`GeometricOptimizers` added as a path dependency** (`deps/GeometricOptimizers`), then subsequently switched to the registered path.

**Tests updated** to reflect the new GO-backed API (cache field names, constructor signatures, retraction types).

### Why

Previously GML duplicated the manifold optimizer logic (Cayley/geodesic retractions, global sections, Adam/Momentum/Gradient on Stiefel/Grassmann) that already lives in `GeometricOptimizers`. The duplication made it hard to keep both packages in sync and prevented users from combining GML networks with GO optimizers directly. This PR eliminates the duplication while keeping backward-compatible `Optimizer(method, nn)` call syntax in GML.

### Known issues fixed in this PR

- `Adam{Float32}` used with a `Float64` network previously caused `NewtonOptimizerCache` to be dispatched for `NamedTuple`-wrapped `StiefelManifold` parameters (because the Adam-specific `OptimizerCache` method requires matching `T`). Fixed by `_adapt_method_to_T`, which converts the optimizer's numeric type to the parameter's element type before calling GO.

---

## Follow-up upstream: [GeometricOptimizers#33](https://github.com/JuliaGNI/GeometricOptimizers.jl/pull/33)

The one piece of optimizer machinery this PR still keeps in GML — `AdamOptimizerWithDecay` — does not need to be kept. GO's `DecayingStatic` line search already implements its schedule factor for factor:

| | GML's `AdamOptimizerWithDecay` | GO's `DecayingStatic` |
|---|---|---|
| decay factor | γ = exp(log(η₂/η₁)/n_epochs) | γ = exp(log(η₂/η₁)/n) |
| step at t | η₁γᵗ | γᵗη₁ |

What GO lacked was only the *name*: its 0.2.0 release split the step size out of the `OptimizerMethod`s, and the bundling of Adam's ρ₁, ρ₂, δ with η₁, η₂, n_epochs went with it. **GO#33** restores that name as a convenience pairing returning the `(algorithm, linesearch)` NamedTuple, so GML can *delete* `src/utils.jl:208-229` rather than port it, and rewrite call sites as:

```julia
Optimizer(x, problem; AdamOptimizerWithDecay(n_epochs)...)
```

GO#33 also adds 33 assertions pinning `step_size(ls, t) ≈ η₁·γ_gmlᵗ` — the equivalence that licenses the deletion — and a docs section separating this *learning-rate* decay from `AdamWithEuclideanDecay`'s *weight* decay, since the two share a word and nothing else.

GO#33 is a **draft** stacked on [GO#29](https://github.com/JuliaGNI/GeometricOptimizers.jl/pull/29), which must merge first. The GML-side deletion therefore lands in a follow-up, not here.

The remaining `src/utils.jl` machinery — `GMLEuclideanState`, `_make_optimizer_cache`, `_make_optimizer_state`, `_tree_optim_step!`, `_leaf_optim_step!` — is slated to move to GO the same way. See `CI_FIX_PLAN.md` §4.


michakraus (commented):
Reviewed the whole diff and exercised the new optimizer path against the package's own architectures. The direction is right — the duplication really was a maintenance problem, and the CHANGELOG is unusually good. But the migration is incomplete in ways CI does not catch: **four of the package's architectures cannot construct an optimizer at all**, momentum diverges, and `AdamOptimizerWithDecay` throws on any manifold weight. Julia 1.10/1.12/1.13 are green because no test covers those combinations — the Documentation job is where they show up.

I have pushed fixes for everything below to this branch; details of each fix are in the commits.

---

## Blocking

### 1. `Optimizer(method, nn)` throws for `LASympNet`, `LinearSymplecticTransformer`, `VolumePreservingFeedForward` and `VolumePreservingTransformer`

`_make_optimizer_cache` routes every leaf that is an `ArrayNamedTuple` into GO's `OptimizerCache`. GO's cache constructors build their `Δg`/`g̃` scratch with `_similar(g)`, and `Base.similar` on GML's structured matrices falls back to a dense `Matrix`, so the constructor's `AT` no longer matches:

```julia
julia> nn = NeuralNetwork(LASympNet(4), Float64);

julia> Optimizer(AdamOptimizer(), nn)
ERROR: MethodError: no method matching AdamCache(
  ::@NamedTuple{weight::SymmetricMatrix{Float64, Vector{Float64}}}, …,
  ::@NamedTuple{weight::Matrix{Float64}})
```

Reproduced for all of `Adam`, `Momentum` and `Gradient`, with `SymmetricMatrix`, `SkewSymMatrix`, `LowerTriangular` and `UpperTriangular` weights. This is the same failure the docs build reports at `sympnet_tutorial.md:101` and `volume_preserving_attention.md:283`.

*Fixed by* type-preserving `Base.similar`/`Base.fill!` on the four array types (mirroring what `stiefel_lie_algebra_horizontal.jl` already did for the Lie-algebra types) plus the elementwise GO bridges they need. I also gathered the bridges that were at the bottom of `stiefel_manifold.jl`/`grassmann_manifold.jl` into a new `src/optimizers/go_bridges.jl` — they are about array arithmetic, not about the manifolds.

### 2. The momentum recursion is wrong, and momentum diverges

`_leaf_optim_step!` updates the state with `p ← p + α∇L`, but `update!(::MomentumCache, …)` forms its direction anticipating `p ← α·p + ∇L`. `_euclidean_update!` has the same defect. This is GeometricOptimizers issue #18 reintroduced on the GML side: for a constant gradient the step grows linearly instead of saturating at `η/(1−α)`.

With `α = 0.9`, `η = 0.1`, the per-step displacement:

| t | 1 | 2 | 3 | 4 | … | 8 |
|---|---|---|---|---|---|---|
| this branch | 0.1 | 0.181 | 0.262 | 0.343 | … | 0.667, still growing |
| `η(1−αᵗ)/(1−α)` | 0.1 | 0.19 | 0.271 | 0.3439 | … | 0.5695 → 1.0 |

No test covers momentum convergence, on a manifold or off it.

*Fixed* in both paths; `test/optimizers/momentum_optimizer.jl` now pins the step sizes against the closed form and checks that Stiefel momentum converges and keeps `Y'Y ≈ I`.

### 3. `AdamOptimizerWithDecay` throws on any manifold parameter

It is not in `_is_go_native_method`, so *every* leaf — manifolds included — falls through to `GMLEuclideanState`, whose constructor is `GMLEuclideanState{T,typeof(x)}(0, zero(x), zero(x))`, and `zero(::StiefelManifold)` is a `StiefelLieAlgHorMatrix`:

```julia
julia> nn = NeuralNetwork(Chain(StiefelLayer(3, 5), Dense(5, 5, tanh)), Float64);

julia> Optimizer(AdamOptimizerWithDecay(10), nn)
ERROR: MethodError: Cannot `convert` an object of type StiefelLieAlgHorMatrix{…}
                    to an object of type StiefelManifold{…}
```

`test/optimizers/optimizer_convergence_tests/adam_with_learning_rate_decay.jl` only uses `Dense`/`ResNetLayer`, so it never reaches this.

*Fixed* by routing the method onto GO's Adam cache and state — it differs from `Adam` only in the step size, which `_current_step_size` already supplies separately. `_make_optimizer_state` now applies `_adapt_method_to_T` the same way `_make_optimizer_cache` does; the asymmetry was what hid this. The decay test now also trains a network with a `StiefelManifold` weight.

### 4. The documentation build fails with ~45 errors

Four root causes, in descending order of blame:

1. **Issue 1 above.** The `sympnet_tutorial`, `volume_preserving_attention` and `volume_preserving_transformer_rigid_body` chains all die at `Optimizer(Adam(type), nn)`; every downstream `@example`, `@setup` and "invalid local link/image" error follows from it. The PNGs are *generated* by the `@setup` blocks that fail — they are not missing files.
2. **Bindings that moved to GO.** `@docs` blocks in `arrays/global_tangent_spaces.md`, `introduction.md` and `manifolds/riemannian_manifolds.md` still name `GlobalSection`, `apply_section`, `global_rep`, `geodesic`, `GradientMethod`, and `GeometricOptimizers` is not in `makedocs(modules = …)`, so neither the blocks nor the `@ref`s pointing at them resolve.
3. **API drift the PR half-fixed.** `optimizer_methods.md` was rewritten but its examples still call `update!(o, o.cache, dx)`, `o.cache.A` and `o.cache.Y`, none of which exist. `parallel_transport.md:124` imports `GeometricMachineLearning: update_section!`. `vec(::StiefelLieAlgHorMatrix)` lost its `@docs` home when `bfgs_optimizer.md` was deleted, which is a `missing_docs` error.
4. **One unrelated failure**, worth folding in while we are here: `volume_preserving_transformer_rigid_body.md` passes `parameters = default_parameters` (the function) to `odeensemble`; current GeometricProblems wants the value.

For the record, the job was red on `main` too — but for a different reason (duplicate BibTeX key `Kraus:2020:GeometricIntegrators`, which this PR's `.bib` change fixes). The build never got far enough there to expose any of the above, so these are this PR's to own.

*Fixed*: `GeometricOptimizers` added to `modules` with `checkdocs_ignored_modules`; the optimizer examples rewritten against `optimization_step!` and the GO accessors; the `parallel_transport` import repointed; `vec(::StiefelLieAlgHorMatrix)` re-homed into `global_tangent_spaces.md`. I also **restored the retraction docstrings**: deleting `retractions.jl` dropped the `geodesic`/`cayley` docstrings and the `## Library Functions` block in `retractions.md` was replaced with a two-line note, so the manual lost that content. They are now attached to the GML-side methods, doctests and all.

---

## Non-blocking (also fixed)

- `Base.zero(::GrassmannManifold)` hardcodes CPU `zeros(T, …)`, dropping the KernelAbstractions backend that both the old code and the `StiefelManifold` method three lines above keep. Restored, and the `StiefelManifold` one now uses `networkbackend` like the rest of the package rather than `KernelAbstractions.get_backend`.
- `_leaf_optim_step!` calls `update_section!` twice per manifold leaf with the same direction. The second one recomputes an `O(N³)` retraction that the first already produced; `_copyto!(section(state), section(cache))` is the same value at `O(N²)`.
- `train!` builds a fresh `GlobalSection(params(nn))` inside the optimization loop — a QR decomposition per manifold weight per step — while `optimize_for_one_epoch!` builds it once per run. Hoisted. Related: `λY` is now effectively an *output* of `optimization_step!` (the live section is `section(state)`), which is worth saying in the docstring; it now is.
- The momentum default step size changed `1e-3 → 1e-2` and Adam's `δ` `3f-7 →` GO's default. Neither is in the CHANGELOG; added under *Changed (breaking)* (keeping the new values, since they follow GO).
- `export SymplecticStiefelLayer` survives although the file defining it is commented out — the companion `SymplecticStiefelManifold` export was removed here. Dropped.
- `.gitignore` gains a bare `*.log`, which is broad enough to swallow real files, and `go_migration_inspection_*.txt`, which is a personal-workflow artifact rather than something the project produces. Narrowed the first to `docs/**/*.log`, dropped the second.
- `AbstractNeuralNetworks.NetworkLoss(nn::NeuralNetwork)` in `losses.jl` is type piracy — the comment above it says so — and nothing in GML or the tests calls it. Removed; the three methods that dispatch on GML's own architectures are fine and stay.
- `Base.fill!` on `SkewSymMatrix`/`StiefelLieAlgHorMatrix`/`GrassmannLieAlgHorMatrix` fills the *storage*, so `fill!(A, 1)` gives a matrix of ±1 and 0 rather than 1. That is the only sensible reading for these types and it is exactly what GO's `fill!(…, NaN)` needs, but it is surprising enough to deserve the comment it now has.
- `version` is still `0.4.8` although `BFGSOptimizer`, `BFGSCache`, `default_optimizer` and `SymplecticStiefelManifold` are gone and `julia` compat moved 1.9 → 1.10. The CHANGELOG's `[Unreleased]` says "targeting 0.5.0"; bumped `Project.toml` to `0.5.0-DEV` to match.

## Not fixed — for you to decide

- **The diff is much harder to review than it needs to be.** `src/GeometricMachineLearning.jl` was re-indented in full (731 changed lines for roughly 30 lines of real change), and `losses.jl`, `utils.jl` and `hnn_loss.jl` were reflowed. The substance is buried. If the formatting is wanted, a separate commit — or a separate PR — would make both halves reviewable.
- **The stated goal is only half met.** GML no longer duplicates the optimizer *methods*, but `stiefel_manifold.jl` and `grassmann_manifold.jl` now carry GML copies of `global_rep`, `geodesic`, `cayley` and `update_section!` that are line-for-line GO's, because GML's `StiefelManifold` does not subtype `GeometricOptimizers.Manifold`. The duplication moved rather than went away. Making GML's manifolds subtype GO's (or having GO's dispatch on an interface) is the change that would actually remove it, and it is worth its own issue.
- **`GMLEuclideanState` is close to dead code, and my fix to issue 3 makes it more so.** `_use_go_cache` fires first, and a layer's parameters are almost always an `ArrayNamedTuple`, so plain Euclidean weights go through GO's cache as well. For Adam that means `AdamCache` (8 arrays: `x, g, δ, Δg, g̃, m₁, m₂, m̃₂`, plus a `GlobalSection`) and `AdamState` (7 more: `x, x̄, g, ḡ, m₁, m₂, m̃₂`, plus a `GlobalSection`) per parameter, against the two the old `AdamCache` held. That is a real memory regression for large Euclidean networks, and the type introduced to avoid it is bypassed. It needs either a genuine reason to exist or removal — I did not want to decide that inside a review.
- **The PR description is out of date**: it places the integration layer in `src/utils.jl` (~250 lines) when it is in `src/optimizers/optimizer.jl`, and points at `CI_FIX_PLAN.md §4`, which is not in the repo.

## Verification

- full suite green, including two new files (`test/optimizers/momentum_optimizer.jl`, `test/optimizers/structured_array_parameters.jl`) and the extended Adam-with-decay test;
- 7 architectures × 4 optimizer methods construct and train for 2 epochs; before, the 12 combinations in issue 1 threw outright, as did `AdamOptimizerWithDecay` on every network with a manifold or triangular weight;
- momentum step sizes match `η(1−αᵗ)/(1−α)` to machine precision, and Stiefel momentum still converges with `Y'Y ≈ I` after the duplicate retraction was removed.

The documentation fixes I could only verify piecewise locally — each rewritten example and doctest runs against the real package, and the bindings all resolve — but a full Documenter run is 3.5 h, so the Documentation and LatexDocs jobs on this push are the real check. I will follow up on whatever they still report.

PR #232, opened 2026-06-24
Fix HDF5 save/load for networks with StiefelManifold and SymmetricMatrix parameters

benedict-96:
## Problem

`AbstractNeuralNetworks.h5save` only handles `AbstractArray` leaves. `PSDLayer` stores a `StiefelManifold` and `LASympNet`'s `LinearLayer` stores a `SymmetricMatrix`, so calling `save` on any network containing these layers threw a `MethodError`.

`changebackend` had the same gap: it only dispatched on `AbstractArray` and `NamedTuple`, so `map_to_cpu` failed for these network types.

## Solution

Adds `src/hdf5_support.jl` with:

- **`h5save`** overloads for `StiefelManifold`, `SymmetricMatrix`, and `SkewSymMatrix`. Each writes a `gml_type` HDF5 attribute so the loader can reconstruct the correct type without ambiguity.
- **`_gml_h5load`** (private) — recursive loader that checks `gml_type` and reconstructs GML types; falls back to a `NamedTuple` for plain groups (matching the existing `AbstractNeuralNetworks` behaviour).
- **`save(h5, nn)` / `save(filename, nn)`** — new dispatches on `NeuralNetwork` that extend `AbstractNeuralNetworks.save` (both open-store and convenience-file variants).
- **`load(NeuralNetwork, h5, arch)` / `load(NeuralNetwork, filename, arch)`** — symmetric extension of `AbstractNeuralNetworks.load`.
- **`changebackend`** overloads for all three special array types, enabling CPU ↔ GPU transfer for any network that contains them.

## Tests (`test/hdf5_support.jl`)

| Test | What it checks |
|---|---|
| `save`/`load` roundtrip: `SymplecticAutoencoder` | `StiefelManifold` survives HDF5 round-trip |
| `save`/`load` roundtrip: `LASympNet` | `SymmetricMatrix` survives HDF5 round-trip |
| `save`/`load` roundtrip: `GSympNet` | plain-array path still works (regression guard) |
| Float32 weight preservation | `eltype` is preserved after reload |
| Open `H5DataStore` API | lower-level `save(h5, nn)` / `load(NeuralNetwork, h5, arch)` |
| `changebackend`: `StiefelManifold`, `SymmetricMatrix`, `SkewSymMatrix` | each type round-trips CPU → CPU |
| `changebackend`: full SAE network | smoke test — output unchanged after backend transfer |

## Motivation

This fix unblocks serialisation and GPU migration of SAE networks with `PSDLayer` blocks, needed for the [SciCade26 pendulum reduced-order modelling demo](https://github.com/JuliaGNI/GeometricMachineLearning.jl/tree/sae-with-highly-nonlinear-decoder).

---

🤖 Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


docs/src/architectures/abstract_neural_networks.md
benedict-96, line 1: The commit message for https://github.com/JuliaGNI/GeometricMachineLearning.jl/pull/232/changes/6b8edecfb35327657d8f8401a6f475658f64c548 should read:
`save` and `load` are exported. There is no need to specify their origin here and this was causing errors before.

benedict-96:
This pr will be set to "draft" because HDF5 support should be made an extension. For `AbstractNeuralNetworks` this was already done here: https://github.com/JuliaGNI/AbstractNeuralNetworks.jl/pull/32

We will do the same in GML and then reopen this pr.

PR #236, opened 2026-08-16
Move the MNIST material to GMLDatasets

michakraus:
`GeometricMachineLearning` is a library for scientific machine learning and had no business pulling an image-dataset package into its documentation build to document itself. The MNIST tutorial **downloaded the data set every time the docs were built** — through an `@setup` block in `mnist_tutorial.md`, and again through `docs/src/tutorials/mnist/Makefile`, which both `Documenter.yml` and `Latex.yml` invoked. `MLDatasets` is now absent from `docs/Project.toml` and `scripts/Project.toml`, and the docs build is offline again.

Everything removed here lands in **[GMLDatasets](https://github.com/JuliaGNI/GMLDatasets)**, together with the MNIST material that was in `GeometricOptimizers`.

## Breaking: removed from `src/`

- **`split_and_flatten` and `onehotbatch`**, with `src/data_loader/mnist_utils.jl` and the unexported index arithmetic behind them (`patch_index`, `within_patch_index`, `index_conversion`).
- **`DataLoader(::AbstractArray{T, 3}, ::AbstractVector)`**, the image-classification constructor. Its own docstring described it as "tailored towards being used with the package MLDatasets.jl", which is what made it the cut. Every other `DataLoader` constructor is unchanged.

  It has to live outside this package in any case: it is the only way to build a `DataLoader` whose input is `Float32` and whose output is `Int`, since the generic tensor–tensor constructor requires a single shared element type.

**No deprecation shims.** GMLDatasets depends on this package, so a forwarding shim here would be a dependency cycle. Porting a script is two lines:

```julia
# before
using GeometricMachineLearning
import MLDatasets
train_x, train_y = MLDatasets.MNIST(split = :train)[:]
dl = DataLoader(train_x, train_y; patch_length = 7)

# after
using GeometricMachineLearning, GMLDatasets
dl = mnist_data_loader(:train; patch_length = 7)
```

**What stayed:** `ClassificationTransformer`, `ClassificationLayer`, `ClassificationTransformerLoss` and `accuracy`. None of them is specific to image data, and they are what GMLDatasets' tutorial trains.

## Also moved out

- `docs/src/tutorials/mnist/` in full, including the pre-trained `.h5` parameters and the TikZ sources for the patch-splitting figure.
- Seven scripts: `transformer_mnist.jl`, `transformer_fashion_mnist.jl`, `transformer_bfgs.jl`, `transformer_analysis.jl`, `autoencoder.jl`, `classifier.jl`, `manifold_based/mnist_grassmann.jl`.
- The MNIST half of `scripts/convert_jld2_to_h5.jl`. Its symplectic-autoencoder and rigid-body halves stayed — those are this package's own.

## Consequences handled

- `ClassificationLayer` and the two `accuracy` methods were documented **only** in the tutorial's `@docs` block, so they are re-homed in `architectures/transformer.md` and `data_loader/data_loader.md`.
- The eight `@ref "MNIST Tutorial"` cross-references become plain links to the GMLDatasets documentation.
- The `mnist_image` target and its three uses are gone from `docs/Makefile`; the "Make MNIST image" steps are gone from both workflows.
- `test/data_loader/optimizer_functor_with_adam.jl` tests the **optimizer**, not MNIST, so it keeps its coverage: it now builds its `DataLoader` from already-patched tensors through the parametric constructor instead of through the removed one.

## Verification

- `Pkg.test()` — full suite green.
- `docs/check_references.jl` — **0 unresolved references**. This is what catches a missed `@ref "MNIST Tutorial"` or an orphaned docstring, and it is the step CI runs before the full build for exactly that reason.
- `docs/` and `scripts/` environments both resolve from scratch without `MLDatasets`.
- `grep -r MLDatasets` over the repository returns nothing outside `CHANGELOG.md` prose.

## Merge order

This should merge **before** the GMLDatasets branch. GMLDatasets resolves `GeometricMachineLearning` from `main`, and until this lands it would pick up a version that still defines `split_and_flatten`, `onehotbatch` and the classification `DataLoader` constructor, and collide with its own.


michakraus (commented):
## Review

The core of this is right, and the removal itself is clean: `git grep` over the branch finds no
dangling reference to `split_and_flatten`, `onehotbatch`, `patch_index`, `within_patch_index`,
`index_conversion`, `MLDatasets` or the removed `DataLoader` constructor outside `CHANGELOG.md`
prose. `docs/check_references.jl` reports 0 unresolved references, and the re-homing of
`ClassificationLayer` and `accuracy` into `architectures/transformer.md` and
`data_loader/data_loader.md` is correct. The rewrite of
`test/data_loader/optimizer_functor_with_adam.jl` reproduces the removed constructor's fields
exactly — `(input_dim, n_patches, n_images, n_classes, 1)` matches the old
`(patch_length^2, number_of_patches, n_params, 10, 1)` — so it keeps the coverage it claims to.

What the PR misses is the *fallout* of deleting `mnist_tutorial.md`. That file was not only a
tutorial: it carried the `@raw latex` chapter introduction for a chapter that still exists, and the
chapter's closing summary lives in a different file and still describes it. I have pushed a
follow-up commit fixing everything below except the blocker.

---

### 🔴 Blocker: every new link 404s

Nine URLs replace the removed `@ref "MNIST Tutorial"` cross-references, and none of them resolve:

| URL | occurrences | status |
|---|---|---|
| `https://github.com/JuliaGNI/GMLDatasets.jl` | `CHANGELOG.md:44`, `src/architectures/transformer_neural_network.jl:6` | **404** |
| `https://juliagni.github.io/GMLDatasets.jl/latest/mnist/mnist_tutorial/` | 7 doc files | **404** |

The repository that exists is **`JuliaGNI/GMLDatasets`** — no `.jl` suffix — and it currently holds
a single `README.md` with GitHub Pages disabled. Note that this PR's own description links
`github.com/JuliaGNI/GMLDatasets` (correct) while the files it adds link `GMLDatasets.jl`
(incorrect), so the two disagree.

The second row matters more than the first, because
`src/architectures/transformer_neural_network.jl:6` is a **shipped docstring** — a dead link there
renders in the manual and in the REPL for every user, not just in a changelog.

Resolution agreed offline: rename `JuliaGNI/GMLDatasets` → `JuliaGNI/GMLDatasets.jl` rather than
change the links. Two things then have to hold before merge:

1. the rename happens (GitHub will redirect the old URL, so nothing else breaks), and
2. GMLDatasets' `deploydocs` sets `devurl = "latest"`, matching `docs/make.jl:361` here — the
   Documenter default is `dev`, which would leave all seven deep links broken.

I have deliberately **not** changed these URLs, so they will stay red until the rename lands.

### 🟠 The LaTeX chapter "Transformers with Structure" lost its introduction

`mnist_tutorial.md` opened with a `@raw latex` block that was not about MNIST — it was the chapter
introduction for the whole chapter:

> In this chapter we discuss examples of improving transformers by imbuing them with structure. We
> show two examples: an example of using the vision transformer … and using the volume-preserving
> transformer to learn the dynamics of a rigid body. At the end we further compare two different
> approaches to realizing the volume-preserving transformer.

With MNIST gone, `_latex_pages` (`docs/make.jl:315`) makes
`tutorials/volume_preserving_transformer_rigid_body.md` the chapter's first page — and that file has
no `@raw latex` intro. It became the only chapter in `_latex_pages` without one; every other
chapter-opening page (`sympnet_tutorial.md`, `symplectic_autoencoder.md`, `grassmann_layer.md`,
`adjusting_the_loss_function.md`) has one.

Fixed by adding an intro to the rigid-body page, rewritten for the two pages the chapter still has.
The `\footnotemark[1]`/`\footnotetext[1]` pair from the old intro is dropped — it justified
excluding the linear symplectic transformer from a *two-example* chapter that no longer has two
examples.

### 🟠 The chapter summary still describes MNIST

`docs/src/tutorials/volume_preserving_attention.md` closes the same chapter with a
`\section*{Chapter Summary}` that still reads:

> The two examples we gave were **(i) enforcing orthogonality constraints for some of the weights in
> a vision transformer** … and (ii) using a volume-preserving transformer … **In the first case the
> network was not able to learn anything if orthogonality constraints were not imposed** …

Example (i) is no longer in the chapter, or in this repository at all. Rewritten around the two
remaining pages. Its `@bibliography` block also still lists `brantner2023generalizing`, which was
the MNIST/optimizer paper and is now cited nowhere in the chapter — dropped.

### 🟡 Minor

- **`.gitignore`** — removing `docs/src/tutorials/mnist/*.pdf` left the comment
  `# PDFs compiled from the .tex sources next to them` with nothing under it. Removed.
- **`docs/src/architectures/transformer.md:60`** — the new entry reads
  `GeometricMachineLearning.ClassificationLayer`, but `ClassificationLayer` **is** exported
  (`src/GeometricMachineLearning.jl:287`). In that block the prefix marks unexported names
  (`GeometricMachineLearning.assign_output_estimate`); `Transformer` and `ClassificationTransformer`
  are bare. Prefix dropped. (`GeometricMachineLearning.accuracy` in `data_loader.md` is correct —
  `accuracy` genuinely is not exported.)
- **`docs/make.jl:315-317`** — deleting `"MNIST",` left the two surviving arguments at the old
  continuation indentation, ~20 columns adrift. Realigned.
- **`test/data_loader/optimizer_functor_with_adam.jl:38`** —
  `ClassificationLayer(dl.input_dim, 10, σ)` hardcodes the class count while
  `create_dummy_classification_data` takes it as a `n_classes` keyword, so the two can silently
  disagree. Now `dl.output_dim`, with `n_classes` threaded through `test_optimization_with_adam`.
- **`CHANGELOG.md`** — two things. The four pre-existing bullets under `### Removed (breaking)` are
  tight; the two new ones were blank-line separated, which makes CommonMark render the *entire*
  list loose. And the long prose block ("All of this moved to GMLDatasets…", the porting snippet,
  "What stayed") was nested under the `DataLoader(::AbstractArray{T, 3}, ::AbstractVector)` bullet
  although it describes the whole move rather than that one constructor. Spacing made consistent and
  the prose hoisted to a lead paragraph for both MNIST entries.
- **`docs/src/data_loader/data_loader.md`** — the PR edits the last line of a file that has no
  trailing newline. Added.

### 🔵 Noted, not changed

- **`accuracy` and `ClassificationTransformer` are now exercised nowhere.** The PR keeps them
  deliberately, but neither appears in `test/` on `main` or on this branch — the MNIST tutorial was
  their only execution path, and it is leaving. (`ClassificationLayer` is fine;
  `test/layers/classification.jl` covers it.) `optimizer_functor_with_adam.jl` already builds a
  suitable classification `DataLoader`, so a smoke test would be a few lines. Happy to add it if you
  want it here rather than in GMLDatasets.
- **The test builds its `DataLoader` through the inner constructor**, `DataLoader{T,…}(input,
  output, …)`, which couples it to field order and count. Using the public tensor–tensor constructor
  (`src/data_loader/data_loader.jl:172`) would force a `Float32` one-hot target instead of `Int`.
  Keeping `Int` is the more faithful stand-in for what GMLDatasets produces, so this looks like the
  right trade-off — just flagging that the coupling is there.
- **Pre-existing, out of scope:** `Plots` in `docs/Project.toml` and `Functors`/`ProfileView` in
  `scripts/Project.toml` are unused, but they were already unused on `main` — this PR did not
  orphan them.

### Verified

- `julia --project=docs docs/check_references.jl` → `0 unresolved reference(s)`, after the fixes.
- `Pkg.test()` → full suite green. (The local `Manifest.toml` had to be regenerated first: it
  pinned GeometricOptimizers 0.3.0 against `Project.toml`'s `0.2.1` compat. Untracked, unrelated to
  this PR.)
- `curl -L` against both GMLDatasets URLs → `404` and `404`.
- Every chapter-opening page in `_latex_pages` now has a `@raw latex` intro.
- The two LaTeX fixes are verified by inspection of `_latex_pages` ordering, **not** by a PDF build
  — `make latex -C docs` takes hours and needs xelatex.

PR #237, opened 2026-08-16
Update the README example to the 0.5 optimizer interface

benedict-96:
The example in the README predates the move of the optimizer machinery to [GeometricOptimizers][go] and no longer runs against `main`.

## Changes

| line | before | after |
|---|---|---|
| optimizer | `Optimizer(AdamOptimizer(), g_nn)` | `Optimizer(Adam(type), g_nn; step_size = 1e-3)` |
| prediction | `Iterate_Sympnet(g_nn, ics; n_points = ...)` | `iterate(g_nn, ics; n_points = ...)` |

- **The optimizer constructor.** The method comes first and the learning rate is no longer part of the method, per the *Changed (breaking)* section of the changelog. `Adam` is constructed with the element type of the parameters, so it reuses the `type` binding already defined a few lines above — which also makes the point that the example is type-generic.
- **`Iterate_Sympnet`.** Not defined in the package any more (`isdefined(GeometricMachineLearning, :Iterate_Sympnet) == false`); the line raised an `UndefVarError` as written. `iterate` is what the SympNet tutorial uses.

One paragraph of prose was added saying where the optimizer methods now come from, since the README is where a reader meets them first and `Adam` no longer being a GML type is otherwise unexplained.

## Verification

Ran the changed calls against the package on Julia 1.12 with a synthetic `q`/`p` dataset — `Optimizer(Adam(Float32), nn; step_size = 1e-3)` constructs (`method::Adam{Float32}`, `step_size = 0.001`), a 3-epoch training run completes, and `iterate(nn, ics; n_points = 200)` returns `(1, 200)` trajectories. Confirmed in the same run that `Iterate_Sympnet`, `default_optimizer` and `BFGSOptimizer` are all undefined and that `AdamOptimizer` survives as an alias.

The CUDA and `Plots` lines are unchanged and were **not** exercised — no GPU here, and neither package is in the project environment.

## Not addressed

The README has no installation section, so the Julia 1.10 floor is not stated anywhere in it. Worth a follow-up if you want it mentioned outside the changelog.

[go]: https://github.com/JuliaGNI/GeometricOptimizers.jl

---

Authored by Claude Opus 5 (Claude Code) at @benedict-96's request. Documentation only — no change under `src/`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ At its core every neural network comprises three components: a neural network ar
 
 Traditionally, physical properties have been encoded into the loss function (PINN approach), but in `GeometricMachineLearning.jl` this is exclusively done through the architectures and the optimizers of the neural network, thus giving theoretical guarantees that these properties are actually preserved.
 
+The optimizer methods themselves (`GradientMethod`, `MomentumMethod`, `Adam`, and the caches, global sections and retractions that go with them) come from [`GeometricOptimizers.jl`](https://github.com/JuliaGNI/GeometricOptimizers.jl) and are re-exported here; `GeometricMachineLearning.jl` supplies the part that is specific to neural networks, i.e. walking the parameter tree and the manifold layers.
+
 Using the package is very straightforward and is very flexible with respect to the device `(CPU, CUDA, Metal, ...)` and the type `(Float16, Float32, Float64, ...)` you want to use. The following is a simple example to learn a SympNet on data coming from a pendulum:
 ```julia
 using GeometricMachineLearning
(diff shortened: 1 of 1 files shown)


michakraus:
Superseded by #241, which carries both of this PR's substantive fixes — `Optimizer(Adam(type), g_nn; step_size = 1e-3)` and `iterate` in place of the undefined `Iterate_Sympnet` — along with the paragraph on where the optimizer methods now come from.

The plotting hunk here was overtaken by #238, which converted the README to CairoMakie on `main`. The API fixes only make sense alongside the rest of 0.5.0, so they travel with that release rather than landing separately.

Closing as superseded, not rejected.

PR #238, opened 2026-08-16
One plotting library, and the bugs that finding it uncovered

michakraus:
> **Stacked on #236.** Base is `move-mnist-material-to-gmldatasets`, so the diff below is only the six commits on top. GitHub will retarget this to `main` when #236 merges.

Three plotting libraries were in use where one would do. CairoMakie was already the house style — every 2D figure used it, and `docs/src/manifolds/manifolds.md` even rendered a full 3D scene with it — while GLMakie survived in six doc pages and Plots in `scripts/` and `hnn/`.

Chasing that down turned up a series of latent bugs, which is most of what this PR is.

## The plotting migration

**The payoff is in CI.** GLMakie needed `xorg-dev mesa-utils xvfb libgl1 freeglut3-dev libx*` installed and both the documentation and LaTeX builds wrapped in `xvfb-run`. None of that is here any more. It also needed `docs/gl_makie_transparent_background_hack.jl`, a `colorbuffer` trick to save a transparent background, which CairoMakie does natively — deleted.

**On the 3D question.** CairoMakie's limitation is real and unchanged: no z-clipping, depth resolved per plot rather than per pixel. It does not bite here. None of the six pages uses interactivity or animation; they are static `Axis3` renders. I rendered the hardest one (`parallel_transport.md`) three ways and compared against the GLMakie output before touching anything. The one visible difference: GLMakie dimmed arrows on the far side of a semi-transparent sphere, CairoMakie draws them at full strength. Geometry, translucency and decorations all survive, so the figures were ported as they stood.

Also worth recording: the `translate!(obj, 0, 0, z)` depth-sorting trick from the CairoMakie docs is **not usable inside an `Axis3`** — it translates in *data* space, so it displaces the geometry instead of reordering the draw.

## Bugs GLMakie had been hiding

`grassmann_layer.md` failed outright once ported, with `BoundsError: attempt to access 0-element Vector{UInt32} at index [0]`, and the cause was not the backend:

- The `Axis3` of `make_point_cloud_arrows` was limited to the graph of the Rosenbrock function while `D_1` sits around `(2, 2, 2)` — so the red cloud and the tail of every arrow were **outside the axis**, and the figure had been showing half of what its caption describes. CairoMakie does not tolerate a scatter clipped away in its entirety.
- The arrows were drawn at `alpha = .01`, relying on GLMakie accumulating thirty overlapping transparent arrows into something visible. CairoMakie composites in order, so they vanished.

Both fixed; the figure now shows the gradients it is captioned for.

## `TrainingParameters` and `train!`

`TrainingParameters(nruns, method)` raised **`UndefVarError: default_optimizer not defined`**. The default was deleted with GML's optimizer layer; the call site was left behind. `TrainingParameters(nn, data)` was worse — it called *both* `default_optimizer()` and `default_integrator(nn, data)`, and neither exists.

Nothing caught it because everything touching `TrainingParameters` lives under `test/train!/`, which `runtests.jl` does not include. The new `test/training_parameters.jl` **is** included.

Two more in `train!`, each making a documented signature unusable:

- **The step size could not be set at all.** It moved onto `Optimizer`, and `train!` built its `Optimizer` without forwarding one, so every run silently took the default.
- **The default training method was `default_method(nn, data)`** while the parameter is named `_data` — so `data` resolved to GML's exported accessor function and the three-argument form always raised.

## The HNN pendulum script

Five generations of drift, each hiding the next: the two-argument `MomentumOptimizer`, the keyword `HamiltonianArchitecture` constructor, `∇H`/`dH` deleted from `pendulum.jl` while `get_data_set` still called them, and `get_data_set` returning a bare `(data, target)` pair where `train!` wants a `TrainingData`.

**It still does not run, and the blocker is in the library**: every HNN training method calls `vectorfield(nn, x, params)` (`hnn_exact_method.jl:6`, `symplectic_euler.jl:13,18`) and no such method exists — `vectorfield` resolves only to GeometricBase's `State` methods. HNN training through `train!` is dead for every method. Noted at the top of the script; separate work.

## Dependencies

Dropped from `[deps]`: `BandedMatrices`, `SparseArrays`, `StatsBase`, `ZygoteRules`. Moved `Printf`, `ChainRulesTestUtils`, `SafeTestsets` to `[extras]`. Dropped `Aqua`, `RungeKutta` and redundant entries; dropped `Documenter` from the test target.

**`Distances` looked equally dead and is not** — `sqeuclidean` is the default distance of every `TrainingMethod`, reached through the module-level `using`. The test suite caught it. It stays, with a comment.

`scripts/Project.toml` lost three dead entries and **gained ten it was using without declaring**; several scripts could not run in that environment. `hnn/Project.toml` likewise.

## `legacy/`

`code_generation` → `legacy/codegen`, `mt_fun` → `legacy/mtk`, `hnn` → `legacy/hnn`. Nothing outside them referenced them. Relative includes rewritten; `hnn_lux_metal.jl`, the only untracked one of the seven `hnn_*.jl` scripts, is committed.

## Corrections to the changelog

The entry claiming GeometricOptimizers' `_BFGS()` replaces GML's removed `BFGSOptimizer` was **wrong** — they are different algorithms, and BFGS training of a network is lost rather than relocated. See the commit for the detail and what bridging it would take. I attempted the bridge and backed it out: GML's `StiefelManifold` is a *different type* from GeometricOptimizers', with unrelated hierarchies, so none of GO's `Manifold` helpers apply.

## Verification

- Full test suite passes.
- The complete documentation builds **without `xvfb`** — which is the proof GLMakie is gone — and `docs/check_references.jl` reports no unresolved references.
- `plot_hnn` exercised end to end on a synthetic Hamiltonian; every figure compared against the previous `docs/build/` output.


Project.toml
```diff
@@ -20,17 +18,12 @@ KernelAbstractions = "63c18a36-062a-441e-b654-da1e3ab1ce7c"
...
-SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
-SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
-StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 SymbolicNeuralNetworks = "aed23131-dcd0-47ca-8090-d21e605652e3"
 Symbolics = "0c5d862f-8b57-4792-8d23-62f2024744c7"
 TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
-ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 
 [weakdeps]
```
michakraus, line 28: **Blocking, fixed in the commit above.** A `[sources]` block sat here:

```toml
[sources]
GeometricOptimizers = {path = "../GeometricOptimizers"}
```

That is local dev state, and it breaks the package for everyone else: `Pkg.instantiate` resolves GeometricOptimizers from a sibling directory that exists on one machine, and the `GeometricOptimizers = "0.2.1"` bound below is bypassed while it is there.

It also contradicted the *Dependencies* entry in the CHANGELOG, which says that entry "is gone now that it is registered, which also makes GML registrable again".

michakraus (commented):
## Review

The substance of this holds up, and I checked the load-bearing claims against `src/` rather than taking them:

- every dropped dependency really is unused — `BandedMatrices` (no `BandedMatrix`, no `bandwidth`), `SparseArrays`, `StatsBase`, `ZygoteRules` (no `@adjoint`, no `@nograd`), `Printf`/`Aqua`/`RungeKutta`/`Documenter` — and `Distances` really is load-bearing through the module-level `using`;
- `_default_step_size` and `Optimizer(method, nn; step_size)` exist, so the `train!` fix dispatches;
- `StandardHamiltonianArchitecture(dim, width, nhidden, activation)` matches `(ninput, ld, ln)`, i.e. the positional rewrite preserves `nhidden = ln, width = ld`;
- `_is_go_native_method(::GeometricOptimizers.OptimizerMethod) = false` — so the corrected BFGS entry is right about where `BFGS` gets routed;
- the new `TrainingData` in `scripts/pendulum.jl` really does land on `DataSymbol{DerivativePhaseSpaceSymbol}`/`SampledData` with `dim == 2`, hence `ExactHnn`. (`DataSymbol(::Tuple)` compares tuples for *equality* first, so it only works because the `⊆` branch catches the arbitrary `Dict` key order — worth knowing, not worth changing here.)
- the Plots→Makie contour transposes (`for y in Y, x in X` → `for x in X, y in Y`) are the right way round, and `px_per_unit = Main.output_type == :html ? 1.5 : 2` matches the existing idiom in `docs/src/manifolds/manifolds.md`.

What it also carries is three blocks of **local Pkg dev state**, and four real bugs. All of it is fixed in the commit pushed on top; details inline.

### Blocking — committed local dev state

This is the one thing here that would have hurt. `Project.toml` gained

```toml
[sources]
GeometricOptimizers = {path = "../GeometricOptimizers"}
```

which makes every consumer — users, CI, the registry — resolve GeometricOptimizers from a sibling directory that does not exist, and bypasses the `GeometricOptimizers = "0.2.1"` bound. It also contradicts the CHANGELOG entry two sections down, which says the `[sources]` entry "is gone now that it is registered, which also makes GML registrable again".

`docs/Project.toml` and `scripts/Project.toml` each gained a `{path = "/Users/mkraus/Datashare/…"}` line. The `scripts/` one is worse than it looks: it sits **inside `[deps]`, duplicating a key that is already there**, and duplicate keys are a TOML parse error — so no script can use that environment at all, which is the opposite of what "declare the ten packages the scripts actually use" set out to achieve. There is a dangling empty `[sources]` header at EOF too.

All three removed. `docs/` keeps its `BrenierTwoFluid` source entry — that one is a genuine `[deps]` package and agrees with what `install_brenier_two_fluid` does; GML itself is developed by `docs/Makefile`, which is what makes the file work at any clone path.

### Bugs

- **`scripts/Script_using_fully_GML/RNN/validation.jl`** calls `axislegend(ax)` twice, either side of `vlines!` — two legends on top of each other, the first missing the `vlines!` entry. The sibling LSTM file has it right.
- **`legacy/hnn/hnn_lux_metal.jl`**, newly committed here, saves to `hnn_lux.png` and clobbers `hnn_lux.jl`'s figure.
- **`legacy/mtk/generate_hnn_mt.jl`** had `../utils/networks.jl` rewritten to `../../utils/networks.jl`, faithfully preserving a path that has never existed — while the file it wants became a *sibling* in this very commit: `legacy/hnn/networks.jl`. This is the one of the "8 already-broken includes" that the move made fixable, and it takes the count to 7.
- **`test/training_parameters.jl`'s step-size testset does not test the step size.** Both branches assert only `length(loss) == 1`, so deleting `step_size` from `train!` leaves it green — which matters, because being the regression test the old bug lacked is the whole reason the file exists. Rewritten to assert an invariant that needs no seed: `train!` recomputes the loss over the *whole* data set after every step, so at `step_size = 0` every entry of the loss array must be the same number. Verified by reverting `Optimizer(m, nn; step_size = step_size)` to `Optimizer(m, nn)` and watching it fail (`all(Bool[1, 0, 0, 0, 0])`).

### Minor

A comment in `scripts/hnn_pendulum.jl` that lost its subject; `[deps]` in `Project.toml` no longer sorted (`Distances` before `ChainRulesCore` — Pkg re-sorts on the next write, so it comes back as noise); `ChainRulesTestUtils` losing its `[compat]` while staying in `[extras]`; `legacy/hnn/README.md` saying `networks.jl` is "gone entirely" when it is in that very directory; `axes` and `size` shadowed as locals in two scripts; `train!`'s docstring not mentioning the new `step_size` and still advertising `GradientOptimizer(1e-2)` and `nruns`; and `plot_hnn` routing *both* panels through `Reverse(:viridis)` when under Plots only the Hamiltonian panel had `rev = true` — `_contour_panel!` takes a `colormap` now, so the error panel looks as it did.

One correction to the description: it is 29 include targets under `legacy/`, not 28 — `hnn_lux_metal.jl` brought one with it — of which 22 now resolve.

### CHANGELOG

The CHANGELOG carried only the BFGS correction, while the PR description carries the whole story. Moved into the file, in its existing shape:

- *Changed (breaking)* — `TrainingParameters`/`TrainingSet` taking the optimizer explicitly, CairoMakie as the only plotting library (with `xvfb` leaving CI as the consequence), and the `legacy/` move.
- *Fixed* — the unsettable step size, the three-argument `train!` that always raised, and why `test/training_parameters.jl` had to be new.
- *Dependencies* — the four dead deps, the `Distances` near-miss, the `[extras]` moves, and the `docs`/`scripts` environments.
- *Open Issues* — two entries for what this uncovered and did not fix:
  - **B6.** HNN training through `train!` is broken for every method: `ExactHnn`, `SEulerA` and `SEulerB` all call `vectorfield(nn, x, params)` and no such method exists. That was recorded only in a comment at the top of `scripts/hnn_pendulum.jl`, which is not where a library defect belongs — and nothing under `test/` covers it, because the HNN methods are exercised only from `test/train!/`.
  - **C9.** The seven remaining `legacy/` includes naming `scripts/data.jl` and `src/training.jl`.

### Comments

With the CHANGELOG carrying the history, the comments no longer need to. A number of the new ones narrated it instead of describing the code — "`mopt` used to default to `default_optimizer()`, which was deleted…", "the step size used to live on the optimizer method… and this call did not follow", "`alpha = .01` was tuned for GLMakie…", "`b{0.4h}` in the Plots layout this replaces", "this was a byte-for-byte copy of `scripts/plots.jl`" — which asks a reader to know what the code looked like before in order to understand what it does now, and dates the moment the next change lands.

All rewritten to state the current arrangement and the reason for it, self-contained, with references where a reference is what is wanted: `scripts/hnn_pendulum.jl`'s note now points at *Open Issues → B6* rather than restating that everything above it was repaired. `legacy/hnn/README.md` went further — it described a `utils/` layout, and the paragraph explaining that the layout was gone was itself the third description of the same thing; the file now just names where each of the three files is, and which one does not exist.

Full suite passes with the fixes applied.

Project.toml
```diff
@@ -5,10 +5,8 @@ authors = ["Michael Kraus <michael.kraus@ipp.mpg.de>"]
 
 [deps]
 AbstractNeuralNetworks = "60874f82-5ada-4c70-bd1c-fa6be7711c8a"
-BandedMatrices = "aae01518-5342-5314-be14-df237901396f"
 ChainRules = "082447d4-558c-5d27-93f4-14fc19e9eca2"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
-ChainRulesTestUtils = "cdddcdb0-9152-4a09-a978-84456f9df70a"
 Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
```
michakraus, line 10: Two small things in this file.

`[deps]` had lost its sort order — `Distances` was placed before `ChainRulesCore`. Pkg re-sorts on the next write, so it would have come back as diff noise in an unrelated PR. Reordered.

And `ChainRulesTestUtils` lost its `[compat]` entry while staying in `[extras]` and the test target, where `SafeTestsets` kept its. Restored `ChainRulesTestUtils = "1"`.

scripts/Project.toml
```diff
@@ -1,23 +1,33 @@
...
+ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
+StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 SymbolicNeuralNetworks = "aed23131-dcd0-47ca-8090-d21e605652e3"
 Symbolics = "0c5d862f-8b57-4792-8d23-62f2024744c7"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
+
+# These scripts run against a development checkout of GeometricMachineLearning. Activate this
```
michakraus, line 31: **Blocking, fixed in the commit above.** This file had

```toml
GeometricMachineLearning = "194d25b2-d3f5-49f0-af24-c124f4aa80cc"
GeometricMachineLearning = {path = "/Users/mkraus/Datashare/Julia/GeometricMachineLearning"}
```

— a `[sources]`-style entry sitting *inside* `[deps]`, duplicating the key above it. Duplicate keys are a TOML parse error, so this environment could not be loaded at all, which is the opposite of what declaring the ten packages the scripts use was for:

```
julia> using TOML; TOML.parsefile("scripts/Project.toml")
ERROR: Parsing error ... duplicate key
```

Removed, along with the dangling empty `[sources]` header at EOF, and replaced with this note.

docs/Project.toml
```diff
@@ -6,12 +6,16 @@ CairoMakie = "13f3f980-e62b-5c42-98c6-ff1f3baf88f0"
...
 GeometricIntegrators = "dcce2d33-59f6-5b8d-9047-0defad88ae06"
 GeometricMachineLearning = "194d25b2-d3f5-49f0-af24-c124f4aa80cc"
 GeometricOptimizers = "fc236c15-5557-4942-aa65-b650f329279e"
 GeometricProblems = "18cb22b4-ad41-5c80-9c5f-710df63fbdc9"
 HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
 LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
-Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
+
+# GeometricMachineLearning itself is deliberately *not* listed here: `docs/Makefile`'s `test_docs`
```
michakraus, line 17: **Blocking, fixed in the commit above.** `[sources]` had gained

```toml
GeometricMachineLearning = {path = "/Users/mkraus/Datashare/Julia/GeometricMachineLearning"}
```

Documenter CI has no `/Users/mkraus`, so the docs build would have died at the first `Pkg.instantiate` — and it was redundant besides, since `docs/Makefile`'s `test_docs` target already runs `Pkg.develop(PackageSpec(path=pwd()))`.

The `BrenierTwoFluid` entry stays: that one is a real `[deps]` package and agrees with what `install_brenier_two_fluid` does.

scripts/Script_using_fully_GML/RNN/validation.jl
```diff
@@ -68,8 +68,11 @@ end
...
 
-plt = plot(t, q1, label="Numeric Integration", size=(1000,600))
-plot!(plt, t, data_matrix[1,:], label="Neural Network")
-vline!(plt, [seq_length*time_step-time_step], color="red",label="Start of Prediction")
+plt = Figure(size = (1000, 600))
+ax = Axis(plt[1, 1])
+lines!(ax, t, q1; label = "Numeric Integration")
+lines!(ax, t, data_matrix[1,:]; label = "Neural Network")
+vlines!(ax, [seq_length*time_step-time_step]; color = :red, label = "Start of Prediction")
+axislegend(ax)
```
michakraus, line 76: `axislegend(ax)` was called here *and* again after the `vlines!` below, so the figure got two legends drawn on top of each other — and the first was missing the "Start of Prediction" entry, since `vlines!` had not run yet.

The sibling `LSTM/validation.jl` has the single call in the right place. Dropped the first one.

legacy/hnn/hnn_lux_metal.jl
```diff
@@ -0,0 +1,116 @@
...
+
+# learned Hamiltonian & vector field
+hnn_est(ξ) = hnn(model, ξ, params, state)
+dhnn_est(ξ) = hnn_vf(model, ξ, params, state)
+
+# plot results
+
+include("../../scripts/plots.jl")
+
+plot_hnn(H, hnn_est, total_loss; filename="hnn_lux_metal.png")
```
michakraus, line 116: This file is new in this PR and wrote `hnn_lux.png` — the same filename `hnn_lux.jl` uses, so whichever ran second clobbered the other's figure. Now `hnn_lux_metal.png`.

legacy/mtk/generate_hnn_mt.jl
```diff
@@ -4,7 +4,7 @@ using ModelingToolkit
 include("common.jl")
 
 #this file contains the function that evaluates the network
-include("../utils/networks.jl")
+include("../hnn/networks.jl")
```
michakraus, line 7: The rewrite here was faithful to a path that has never existed: `../utils/networks.jl` from `mt_fun/` and `../../utils/networks.jl` from `legacy/mtk/` both point at `<root>/utils/networks.jl`, and there is no `utils/` in this repository.

But the file it wants became a **sibling** in this very commit — `hnn/networks.jl` moved to `legacy/hnn/networks.jl`. `../hnn/networks.jl` resolves. This is the one of the eight already-broken includes that the move made fixable; fixing it takes the count to seven, which is now recorded as *Open Issues → C9*.

test/training_parameters.jl
```diff
@@ -0,0 +1,55 @@
...
+# `train!` recomputes the loss over the *whole* data set after every step, so the value it stores
+# does not depend on which batch was drawn. That gives an assertion needing no seed: at
+# `step_size = 0` no parameter can move, hence every entry of the loss array is the same number.
+@testset "train! forwards a step size" begin
+    m = BasicSympNet()
+    o = GradientOptimizer()
+    ntraining = 5
+
+    nn_frozen = NeuralNetwork(GSympNet(2; n_layers = 2), Float64)
+    loss_frozen = train!(nn_frozen, tra_ps_data, o, m; ntraining, batch_size = (1, 2), step_size = 0.0)
```
michakraus, line 47: This testset did not test what it is named for. Both branches asserted only `length(loss) == 1`, so removing the `step_size` keyword from `train!` left it green — `step_size = 1e-8` was simply ignored and nothing noticed. That matters more here than it would elsewhere: being the regression test the bug lacked is the entire reason this file exists.

Rewritten around an invariant that needs no seed. `train!` recomputes the loss over the *whole* data set after every step, so the stored value does not depend on which batch was drawn — which means at `step_size = 0` no parameter can move and every entry of the loss array is the same number.

Checked by reverting `Optimizer(m, nn; step_size = step_size)` to `Optimizer(m, nn)`: `Evaluated: all(Bool[1, 0, 0, 0, 0])`.

scripts/hnn_pendulum.jl
```diff
@@ -19,21 +27,25 @@ const ninput = 2
...
+# rather than to the method, so it is passed to `train!` below.
+#opt = GradientOptimizer()
+opt = MomentumOptimizer(0.5)
 
 # create HNN
-hnn = HamiltonianArchitecture(ninput; nhidden = ln, width = ld)
+# The constructor takes `(dim, width, nhidden, activation)` positionally.
+# `StandardHamiltonianArchitecture` is named explicitly, because calling the abstract
+# `HamiltonianArchitecture` warns and defaults to it.
+hnn = StandardHamiltonianArchitecture(ninput, ld, ln)
```
michakraus, line 39: The comment above this had lost its subject — it read "# is named explicitly, because …" — and the line before it had trailing whitespace. Fixed, and reworded to describe the signature rather than how it changed.

The call itself is right, for the record: `StandardHamiltonianArchitecture(dim, width, nhidden, activation)` at `src/architectures/hamiltonian_neural_network.jl:37`, so `(ninput, ld, ln)` preserves the old `nhidden = ln, width = ld`.

legacy/hnn/README.md
```diff
@@ -1,12 +1,17 @@
 
 # Simple Julia Models for Learning Hamiltonian Dynamics
 
+> Legacy. These scripts predate the current `GeometricMachineLearning` API and are kept for
```
michakraus, line 4: This file had the same problem twice over: the body documented a `utils/` layout that does not exist, and the note added here to say the layout was gone was a third description of the same thing — one of which claimed `utils/networks.jl` was "gone entirely" when `networks.jl` is in this very directory and both `hnn_simple.jl` and `legacy/mtk/generate_hnn_mt.jl` include it.

Rewritten to just name where each of the three files is, and which one is nowhere.

scripts/symplectic_autoencoders/plot_waves.jl
```diff
@@ -28,7 +28,10 @@ function plot_curves(number_of_params::Int=3, time_instances::Int=3)
 
     Ω = -0.5:(1 / (N - 1)):0.5
 
-    plot_object = plot(; layout=(time_instances,1))
+    fig = Figure()
+    # one row per time instance, stacked vertically
+    axs = [Axis(fig[i, 1]; title = "", titlealign = :left, titlesize = 12)
```
michakraus, line 33: `axes` shadowed `Base.axes` for the rest of the function. Nothing in `plot_curves` calls it, so it was latent rather than broken, but it is the kind of thing that bites whoever edits the function next. → `axs`.

scripts/ensemblesolution/plots.jl
```diff
@@ -1,38 +1,48 @@
...
+using CairoMakie
 using LaTeXStrings
 using StatsBase
 
+# Each plot comes in two forms, because a `Figure` cannot be nested inside another one: a `plot_*!`
+# that draws into a grid position of an existing figure and returns its axis, and a `plot_*` that
+# wraps that in a figure of its own. `plot_result`, which composes four panels, uses the former;
+# call the latter to get a single plot on its own.
+"Run `f` on the single cell of a new figure of the given size, and return that figure."
+function _standalone(f, figsize)
```
michakraus, line 10: `size` shadowed `Base.size` inside `_standalone`. → `figsize`.

src/training/train.jl
```diff
@@ -20,19 +20,21 @@ Perform a training of a neural networks on data using given method a training Me
...
+- `training_method` : specify the loss function used
+- `ntraining` : number of iterations through the process
 - `batch_size` : size of batch of data used for each step
+- `step_size` : the learning rate. It is a property of the [`Optimizer`](@ref) rather than of the
+  optimization method, and defaults to `_default_step_size(optimizer)`
 
 """
 
-function train!(nn::AbstractNeuralNetwork, _data::AbstractTrainingData, m::OptimizerMethod, method = default_method(nn, data); ntraining = DEFAULT_NRUNS, batch_size = missing, showprogress::Bool = false, timer::Bool = false)
+function train!(nn::AbstractNeuralNetwork, _data::AbstractTrainingData, m::OptimizerMethod, method = default_method(nn, _data); ntraining = DEFAULT_NRUNS, batch_size = missing, showprogress::Bool = false, timer::Bool = false, step_size::Real = _default_step_size(m))
```
michakraus, line 37: The new `step_size` keyword was not in the docstring above, which also still advertised two things that are no longer true: `optimizer = GradientOptimizer(1e-2)` — the one-argument form went away with the very migration this keyword exists because of — and `nruns`, where the keyword is `ntraining`. All three updated.

The `_data` fix on this line is right and worth stating plainly: the default argument had been `default_method(nn, data)`, and `data` is GML's *exported accessor function* of that name, so the three-argument form raised a `MethodError` for everyone, always. That is now in the CHANGELOG under *Fixed*.

README.md
```diff
@@ -54,8 +54,12 @@ g_loss_array = g_opt(g_nn, dl, Batch(batch_size), nepochs)
 ics = (q=qp_data.q[:,1], p=qp_data.p[:,1])
 const steps_to_plot = 200
 g_trajectory = Iterate_Sympnet(g_nn, ics; n_points = steps_to_plot)
-p2 = plot(qp_data.q'[1:steps_to_plot], qp_data.p'[1:steps_to_plot], label="training data")
-plot!(p2, g_trajectory.q', g_trajectory.p', label="G Sympnet")
+fig = Figure()
+ax = Axis(fig[1, 1]; xlabel = "q", ylabel = "p")
+lines!(ax, vec(qp_data.q')[1:steps_to_plot], vec(qp_data.p')[1:steps_to_plot]; label = "training data")
```
michakraus, line 59: `vec` on the next line but not this one. Both work — linear-indexing an adjoint with a range returns a `Vector` — but the asymmetry reads like one of them is doing something the other is not. Made them the same shape.

scripts/plots.jl
```diff
@@ -1,37 +1,54 @@
+using CairoMakie
 using LaTeXStrings
-using Plots
+
+# Fill colours for the contour panels: the Hamiltonian panels use a reversed viridis and the error
+# panel the unreversed one, so that the two read differently at a glance.
+const _reversed_colormap = Reverse(:viridis)
```
michakraus, line 6: Both panels of `plot_hnn` went through `_contour_panel!` and therefore `Reverse(:viridis)`. Under Plots only `plt_cnt` asked for `cgrad(:default, rev = true)`; `plt_err` was a plain `contourf` on the default gradient. Small, but a change to what the figure looks like that fell out of the refactor rather than being chosen.

`_contour_panel!` takes a `colormap` keyword now, defaulting to the reversed map, and the error panel passes the unreversed one.

PR #247, opened 2026-08-23
Remove the recurrent and LSTM architectures

michakraus:
Removes `RecurrentNeuralNetwork` and `LSTMNeuralNetwork`, along with `src/architectures/recurrent_neural_network.jl`, `src/architectures/LSTM_neural_network.jl` and the driver scripts under `scripts/Script_using_fully_GML/{RNN,LSTM}/`.

## They were already unusable

`Chain(::RecurrentNeuralNetwork)` returned an `AbstractNeuralNetworks.GridCell` rather than a `Chain`, and the cells it held still defined the pre-0.6 `initialparameters(cell, backend, T; init, rng)` signature — so constructing a `NeuralNetwork` from either architecture errored out. Nothing in the test suite referenced either one, which is why nothing noticed.

## Why now

AbstractNeuralNetworks 0.7 deleted `src/cells/` outright — `Recurrent`, `LSTM`, `GRU`, `IdentityCell` and `GridCell` ([AbstractNeuralNetworks.jl#34](https://github.com/JuliaGNI/AbstractNeuralNetworks.jl/pull/34)) — so these imports at the top of `GeometricMachineLearning.jl` have to go regardless of what happens to the architectures that used them:

```julia
import AbstractNeuralNetworks: Architecture, Model, AbstractExplicitLayer,
                               AbstractExplicitCell, ...
import AbstractNeuralNetworks: Chain, GridCell
import AbstractNeuralNetworks: Dense, Linear, Recurrent
```

Doing it here, against `main` and the current AbstractNeuralNetworks 0.6.4, keeps the removal independent of the 0.7 bump — #246 then only has to be a dependency change, not a dependency change carrying a behavioural one.

Concretely: #246 as it stands still imports all three removed names, and loading it against 0.7.0 emits

```
WARNING: Imported binding AbstractNeuralNetworks.AbstractExplicitCell was undeclared at import time
WARNING: Imported binding AbstractNeuralNetworks.GridCell was undeclared at import time
WARNING: Imported binding AbstractNeuralNetworks.Recurrent was undeclared at import time
```

Julia downgrades those to warnings so the module still loads, but they should not survive the merge. Once this lands on `main`, #246 picks it up on its next merge from main.

## Docs

The footnote in `docs/src/architectures/neural_network_integrators.md` had flagged the LSTM implementation as likely to be deprecated since it was written; it now says it is gone. The transformer-derived architectures are what GML maintains for time series.

## Verification

- `Pkg.test()` on this branch against AbstractNeuralNetworks 0.6.4 — **56 testsets, zero failures, zero errors**.
- Loads with no undeclared-binding warnings, and `RecurrentNeuralNetwork`/`LSTMNeuralNetwork` are no longer defined.
- `grep -rniE "RecurrentNeuralNetwork|LSTMNeuralNetwork|GridCell|AbstractExplicitCell|IdentityCell"` over `*.jl`/`*.md`/`*.toml` returns nothing outside the CHANGELOG entry documenting the removal.

Separately confirmed while preparing this: with this removal applied on top of #246's branch, **GML's full suite also passes against the released AbstractNeuralNetworks 0.7.0** (with a dev'd SymbolicNeuralNetworks, which needed its own `QPTOAT` → `ArrayOrNamedTuple` fix), zero failures.


michakraus (commented):
Reviewed at `30f47127`. **The removal is correct and complete** — I could not find a way for it to break anything, and I checked the premise rather than taking it on trust. Two minor cleanups, both now pushed as `df89a6e6`.

## What I verified

**The two architectures really were unconstructible.** Against `main` + AbstractNeuralNetworks 0.6.4:

```julia
julia> Chain(RecurrentNeuralNetwork(4, 4, (1,1)))
GridCell{1, 1, Matrix{Any}, Vector{Vector{Float64}}}

julia> NeuralNetwork(RecurrentNeuralNetwork(4, 4, (1,1)), CPU(), Float32)
ERROR: initialparameters not implemented for model type GridCell{1, 1, Matrix{Any}, Vector{Vector{Float64}}}
```

Same error for `LSTMNeuralNetwork(4, (1,1))`. So the description is accurate: nothing could have been depending on these, and no deprecation shim is owed.

**No dangling references.** `git grep -l -in "RecurrentNeuralNetwork|LSTMNeuralNetwork|GridCell|AbstractExplicitCell|IdentityCell"` across the whole tree — every file type, not just `.jl`/`.md`/`.toml` — returns `CHANGELOG.md` and nothing else: the new removal entry plus the historical 0.3.0 line. `src/`, `test/`, `docs/`, `scripts/`, `legacy/` and `.github/` are clean, and `docs/make.jl` never referenced them.

**The import removal is forced, not merely tidy.** `src/cells/{abstract,grid,gru,identity,lstm,recurrent}.jl` are present in AbstractNeuralNetworks 0.6.4 and absent from 0.7.0, so `AbstractExplicitCell`, `GridCell` and `Recurrent` had to go either way. None of the three was ever re-exported by GML, so `RecurrentNeuralNetwork` and `LSTMNeuralNetwork` are the only public names this drops.

## Findings

### 1. The `IdentityActivation` / `ZeroVector` import is left dead — minor

`src/GeometricMachineLearning.jl:71`

```julia
import AbstractNeuralNetworks: IdentityActivation, ZeroVector
```

`Chain(::RecurrentNeuralNetwork)` was the only thing in `src/` that named `IdentityActivation`, so this PR is what makes it dead. `ZeroVector` was dead already — its last `src/` use moved to `legacy/` in `ec8e8fa5`.

Worth saying why this is a nit and not a bug, since it cuts both ways:

- It is **not** a source of undeclared-binding warnings under 0.7 — both names survive there (`ZeroVector` in `src/utils/zero_vector.jl`, `IdentityActivation` still exported), so leaving it would not have undone what the PR set out to do.
- Removing it is **not** observable either. `IdentityActivation` is an AbstractNeuralNetworks export and line 3 is a blanket `using AbstractNeuralNetworks`, so `GeometricMachineLearning.IdentityActivation` still resolves without the explicit import. The two `scripts/loss/write_loss_*.jl` occurrences spell it fully qualified as `AbstractNeuralNetworks.IdentityActivation` and never depended on the line.

So: redundant, and this PR is the one in a position to notice. Deleted.

### 2. The reworded footnote repeats the sentence it hangs off — minor

`docs/src/architectures/neural_network_integrators.md:42`

The body already ends "…we almost always use a transformer-derived architecture when dealing with time series[^2]", and the new footnote then said "It was removed in favour of the transformer-derived architectures, which are the ones maintained for time series here" — the same claim twice, one line apart. Cut to the fact the footnote exists to record:

```markdown
[^2]: `GeometricMachineLearning` used to ship an LSTM implementation. It has been removed.
```

## Checked and dismissed

- **Trailing whitespace** on the footnote line — the entire file is written that way; not this PR's business.
- **`Project.toml` still at `0.5.0`** — this repo bumps at release, and `## [Unreleased]` already carries other breaking entries.
- **Open Issues C8** ("`scripts/` has been dead since SymbolicNeuralNetworks 0.2") — deleting six of those scripts does not close it; the entry still stands as written.
- **C10** (ten undefined exported names) — unaffected; both removed names were defined.

## Verification of the follow-up

- `Pkg.test()` on `df89a6e6`: **56 testsets, zero failures, zero errors** (twice, plus the pre-push hook run).
- Loads with no undeclared-binding warnings; `isdefined` is `false` for `RecurrentNeuralNetwork`, `LSTMNeuralNetwork` and `ZeroVector`, and `IdentityActivation` still resolves through the blanket `using`, as above.

PR #250, opened 2026-08-24
Hand the structured types' backend and section methods back to their owner

michakraus:
> [!NOTE]
> **Both preconditions are met.** `GeometricOptimizers` 0.5.0 is in the General registry and #249 is
> merged, so this resolves and the suite passes with nothing dev'd — see *Update* at the bottom. The
> release matrix (1.10 / 1.12 / 1.13) and the docs build are green on all three platforms. The three
> **nightly** jobs are red for a Julia compiler bug (`Internal error: during type inference`, in
> `hamiltonian_neural_network_tests.jl`) that fails the same three jobs on `main`
> ([run 32801407865](https://github.com/JuliaGNI/GeometricMachineLearning.jl/actions/runs/32801407865)).

## What moves, and why it was wrong here

Two piracies of the same shape. `changebackend` is `AbstractNeuralNetworks`' generic;
`GlobalSection` is `GeometricOptimizers`'. `StiefelManifold`, `SymmetricMatrix` and the rest are
`GeometricOptimizers`' types; `NetworkParameters` is `NeuralNetworkParameters`'. Each method pairs a
generic from one package with a type from another, and **this package owns neither side of either
pair** — so two packages doing the same thing could silently disagree, which is the failure mode the
0.6.0 `h5save` removal was also about.

`GeometricOptimizers` 0.5 defines both, where one owned argument sits on each side.

## Two bugs that go with them

- The five `changebackend` methods lived inside the **HDF5 extension**, which has nothing to do with
  moving arrays between devices. So `changebackend(GPU(), nn)` on a network with a manifold weight was
  a `MethodError` **unless HDF5 happened to be loaded**.
- They covered five types and missed three: `GrassmannManifold` and neither horizontal lift had a
  method. Upstream covers every family in one, because it dispatches on `Manifold`,
  `VectorStorageMatrix` and `AbstractLieAlgHorMatrix` and delegates to the leaf protocol rather than
  enumerating types.

## Verification

The tests needed no rewriting: they import `changebackend` from `AbstractNeuralNetworks`, so
upstream's methods answer by dispatch and the five per-type testsets plus the whole-network smoke
test keep their meaning — which is the check that the move is transparent. Nothing under `src/`
referenced `changebackend` at all. They have since **moved** to `test/changebackend_tests.jl`, since
coverage of a device transfer has no business being reachable only through the HDF5 testset; see the
review below.

Locally, with the upstream branch `Pkg.develop`ed in, the suite passes. Once
`GeometricOptimizers` 0.5.0 is registered this should go green unchanged.

The compat floor moves to `GeometricOptimizers = "0.5"`, which is what makes this depend on that
release rather than merely prefer it.

---

## Update: the other two bounds (commit `fa504e5`)

`GeometricOptimizers` 0.5.0 is now registered, and two more bounds have to move with it — this environment does not resolve until all three do:

```
GeometricOptimizers     = "0.4.1"  →  "0.5"
NeuralNetworkParameters = "0.1"    →  "0.2"
SymbolicNeuralNetworks  = "0.6"    →  "0.7"
```

`GeometricOptimizers` 0.5.0 drops the `ParameterHandling` shim and lets `NeuralNetworkParameters` do the flattening, so the 0.2 container is required rather than preferred — it carries the element type of its leaves as `NetworkParameters{T}`'s first type parameter, which is what lets a parameter set be an `OptimizerSolution{T}` there.

`SymbolicNeuralNetworks` 0.7.0 is needed because 0.6.0 permits only the 0.1 container and so cannot coexist with `GeometricOptimizers` 0.5.

No further code changes: nothing here calls what 0.5.0 removed (`apply_toNT`, the pirated `ParameterHandling` methods on Base types, or `NeuralNetworkParametersExt`, now `src/parameter_protocol.jl` upstream).

### Verification

The suite now passes with **nothing dev'd** — resolved from the registry at `GeometricOptimizers` 0.5.0, `NeuralNetworkParameters` 0.2.1, `SymbolicNeuralNetworks` 0.7.0 and `AbstractNeuralNetworks` 0.7.1 — which supersedes the "with the upstream branch `Pkg.develop`ed in" note above. The `pre-push` hook ran it again before this push.

If CI fails in `Pkg.resolve` rather than in a test, check whether the runner's cached General clone is stale before reading the compat bounds: `registry failed to rebase on origin/master` has caused exactly this failure repeatedly, both in CI and locally.

---

## Review follow-ups (commits `75bf10fc`, `79bafb26`, `90b6fb52`)

1. **Four scratch scripts were staged by accident** — `scripts/{enzyme,harmonic_oscillator_sympnet,sae_script2,zygote}.jl`,
   298 of the original 317 added lines, unrelated to this change. Deleted.
2. **The `changebackend` testsets moved out of `test/hdf5_support.jl`** into
   `test/changebackend_tests.jl`, ungated, and now also cover `GrassmannManifold` and both
   horizontal lifts, plus an assertion on which module answers the call.
3. **The changelog** now records all three compat floors with the reason for each, and counts three
   missing families rather than two.

Still open, deliberately: `Project.toml` says `version = "0.6.0"` while the changelog documents an
unreleased `0.7.0` — that bump wants its own commit before the tag.


michakraus (commented):
## Review

The substance of this PR is right, and I checked it against the installed upstream rather than
taking the claims on trust:

- `changebackend` — `GeometricOptimizers` 0.5.0 defines **one** method in
  `ext/AbstractNeuralNetworksExt.jl` on `Union{Manifold, VectorStorageMatrix, AbstractLieAlgHorMatrix}`,
  delegating to `NeuralNetworkParameters.mapstorage`. That extension is keyed on
  `AbstractNeuralNetworks`, which is a **hard** dependency here, so it is always loaded — the
  "`MethodError` unless HDF5 happened to be loaded" bug really is fixed, not merely relocated.
- `GlobalSection(::NetworkParameters)` — `src/global_sections/global_sections.jl:42` upstream, in
  `src/` and not an extension (`NeuralNetworkParameters` is a hard dep there), character-for-character
  the deleted method and returning the same plain `NamedTuple` tree. The live call sites keep
  working: `src/training/train.jl:65`, `src/data_loader/optimize.jl:91`,
  `docs/src/tutorials/grassmann_layer.md:111`, and eight test files that pass a `NetworkParameters`.
- No dead imports were left behind in either touched file — `NeuralNetworkBackend` and `Architecture`
  are still used four times each in `HDF5Ext`, and `params`/`NetworkParameters` are still used in
  `optimizer.jl`. `apply_toNT`, `NeuralNetworkParametersExt` and the `ParameterHandling` piracy are
  indeed gone upstream, and nothing here reaches for them.
- The three floors agree with what `GeometricOptimizers` 0.5.0 itself requires
  (`AbstractNeuralNetworks = "0.7"`, `NeuralNetworkParameters = "0.2"`).

**On CI:** the three nightly failures are **not** this PR's. They are a Julia compiler bug —
`Internal error: during type inference` / `TypeError(func=:typeassert, ...)`, surfacing in
`hamiltonian_neural_network_tests.jl` — and the same three jobs fail on `main`
([run 32801407865](https://github.com/JuliaGNI/GeometricMachineLearning.jl/actions/runs/32801407865)).
1.10 / 1.12 / 1.13 and Documentation are green on all three platforms, which is the matrix that
matters for a compat-floor change.

Six findings. The first five are fixed in the three follow-up commits below.

---

### 1. Four scratch scripts committed by accident (major)

`95ff9166` also added `scripts/enzyme.jl`, `scripts/harmonic_oscillator_sympnet.jl`,
`scripts/sae_script2.jl` and `scripts/zygote.jl` — **298 of the PR's 317 added lines**, none of them
related to handing the structured types back, and none mentioned in the description. They are
scratch: `# todo` markers, a `mkpath("animations")` side effect, a 500-frame PDF render loop,
`"../docs/src/tutorials/sae_parameters.jld2"`. Two cannot run at all — `Enzyme` is not in
`scripts/Project.toml`, and `sae_script2.jl` spells the container the 0.1 way
(`NeuralNetworkParameters{Tuple(...)}(ps)`), which *this PR's own floor bump* makes wrong.

The same four files have been staged by accident twice before (`451100cb` "Drop four scratch scripts
staged by accident", `dfa5b55a` "Take the scratch scripts back out of the repository"). Worth a
`git add -p` habit or a `scripts/` entry in `.gitignore` for the ones that are pure scratch.

**Fixed:** deleted.

### 2. The `changebackend` tests were still gated on HDF5 (medium)

The description argues that leaving `test/hdf5_support.jl` untouched is the check that the move is
transparent, and as a one-off that is a fair check to have run. But keeping the six testsets there
preserves on the test side exactly the mistake the PR fixes on the source side: `changebackend` has
nothing to do with HDF5, so its coverage must not be reachable only through the HDF5 testset.

**Fixed:** moved to `test/changebackend_tests.jl`, registered next to the `map_to_cpu` entry in
`runtests.jl`, nothing gating it. The move also picks up the three families that had no method here —
`GrassmannManifold` and both horizontal lifts — and asserts the *module* that answers the call, so
"these methods are no longer ours" is a test rather than a sentence. Upstream's
`test/changebackend.jl` stays the thorough test of the eight families; what remains here is the
integration check plus the whole-`NeuralNetwork` case upstream cannot reach.

### 3. The changelog undercounted what was missing (minor)

"Upstream covers the horizontal lifts too, which were missing here" — three families were missing,
not two. Upstream dispatches on `Manifold`, so `GrassmannManifold` had no method here either. (The
description says "missed three" and then names only the two lifts.)

**Fixed** in `CHANGELOG.md`.

### 4. The changelog recorded one of the three floors (minor)

`fa504e54` moved `NeuralNetworkParameters` to 0.2 and `SymbolicNeuralNetworks` to 0.7 as well, and
the environment does not resolve until all three move — but the changelog named only
`GeometricOptimizers` 0.5. For a release whose subject *is* which package owns what, all three
belong there, with the reason for each.

**Fixed:** the "Requires" paragraph now carries all three and why.

### 5. The description's banner is stale (minor)

The `[!IMPORTANT]` block still reads "Draft: blocked on a `GeometricOptimizers` 0.5.0 release" and
"Stacked on #249, which should merge first". #249 merged on 24 Aug, 0.5.0 is registered, this is not
a draft, and the release matrix is green. A reviewer reading top-down is told not to expect green CI
when it is green.

**Fixed:** description updated.

### 6. Two things left alone deliberately

- **`Project.toml` still says `version = "0.6.0"`** while the changelog documents an unreleased
  `## [0.7.0]` containing breaking removals. Release bookkeeping, and the repo has both conventions
  (`882e2d1b` bumped to `0.5.0-DEV` right after a release; `1d8e1b19` bumped along with the breaking
  work) — so it wants its own commit, not this one. Flagging it so it does not get lost before the
  tag.
- **`docs/inventories/GeometricOptimizers.toml` still declares `version = "0.4.0"`.** It is only the
  *fallback* for `InterLinks` — the published URL takes precedence, see `docs/make.jl:15-32` — and
  the docs build passes, so this is a nit rather than a break. Regenerating it wants a current 0.5
  docs build; the `objects.inv` in the sibling checkout is from 23 Aug and of uncertain vintage, so a
  blind regeneration would be worse than leaving it.
- Minor: `HDF5 = "0.16, 0.17"` is wider than upstream's `"0.17"`. Dead width, not a conflict.

PR #251, opened 2026-08-24
Let SymplecticEuler say how it meets the layerwise seam

### LINKED ISSUE #245: SymbolicPullback build time grows multiplicatively with n_integrators
Reported by @benedict-96 in review of #207 ("massive performance issues" with the `SymbolicPullback`). Measured; the summary is that the *per-call* performance is excellent and the *build* is what explodes, multiplicatively in `n_integrators`.

## Per call, the symbolic pullback is what you want

`dim = 4`, `nhidden = 1`, batch 32, comparing `SymbolicPullback(nn, ParametricLoss(), μ)` with `ZygotePullback(ParametricLoss())`:

| network | params | `ZygotePullback` | `SymbolicPullback` |
|---|---|---|---|
| `dim=2, width=2, n_int=1` | 32 | 10.4 ms | **0.1 ms** |
| `dim=2, width=4, n_int=1` | 80 | 23.6 ms | **0.1 ms** |
| `dim=4, width=4, n_int=1` | 88 | 33.6 ms | **0.1 ms** |
| `dim=4, width=4, n_int=2` | 176 | 140.1 ms | — (never builds) |

The symbolic call is flat at 0.1 ms while Zygote degrades from 10 ms to 140 ms — a 100–1000× advantage that grows with the network. So this is worth fixing rather than abandoning.

## The build explodes with `n_integrators`

Phase-by-phase, `dim = 4, width = 4, nhidden = 1`:

| phase | `n_integrators = 1` | `n_integrators = 2` |
|---|---|---|
| symbolic variables | 0.08 s | 0.01 s |
| **build the symbolic loss** | 0.87 s, **3.4 ⋅ 10⁵ chars** | 0.03 s, **1.4 ⋅ 10⁹ chars** |
| `symbolic_differentials` | 0.03 s | 0.01 s |
| `symbolic_derivative` | 0.66 s, **1.5 ⋅ 10⁸ chars** | never finishes |
| `build_nn_function` | 1.34 s | — |

Two observations:

1. **The loss expression grows ~4,100× when `n_integrators` goes 1 → 2**, before any differentiation. At two integrators it is already 1.4 billion characters; the process passes 8 GB and does not return.
2. **Differentiating multiplies by a further ~440×** (3.4 ⋅ 10⁵ → 1.5 ⋅ 10⁸ at one integrator).

## Cause

Each `SymplecticEuler` layer's forward pass calls the executable gradient that `build_gradient` produced for its energy network. Traced on symbolic `Num`s, that call *inlines the whole gradient expression*. Stacking integrators inlines that expression inside itself, so the loss grows multiplicatively in the depth of the chain rather than additively. `cse = true` in `build_nn_function` cannot help: it runs at code generation, long after the expression has been materialised.

## Suggested fix

Do not trace the inlined chain. Compose the pullback layer by layer — each `SymplecticEuler` has already built the gradient of its own energy network, so the chain rule can reuse those instead of re-expanding them. That keeps the expression linear in depth.

This is upstream-shaped work: it needs either a layerwise pullback protocol in `SymbolicNeuralNetworks`, or a `build_nn_function` that can take a composition without inlining it. Related: #243, which tracks the type piracy in the same area.

## For now

`SymbolicPullback(nn, ::ParametricLoss, μ)` throws an `ArgumentError` for `n_integrators > 1` rather than appearing to hang (GML #207). `n_integrators = 1` builds in ≈1.4 s and is fully supported. The default training path uses `ZygotePullback` and is unaffected.

michakraus:
Resolves #245, given [SymbolicNeuralNetworks #56](https://github.com/JuliaGNI/SymbolicNeuralNetworks.jl/pull/56).

> **Base branch.** This targets `pghnn-rebase`, the rebased form of #207's work, because that is where `SymplecticEuler` lives — on `main` the `GeneralizedHamiltonianArchitecture` constructor is still `error("GHNN still has to be implemented!")`. #207 and its branch are untouched.

## What this does

`SymbolicNeuralNetworks` composes the pullback of a chain layer by layer, putting *fresh* symbolic variables between two layers instead of inlining one layer's expression into the next — which is what keeps the symbolic material a sum over layers rather than a product. Its seam was a plain vector, so it assumed every layer maps an array to an array.

A `SymplecticEuler` built with `return_parameters = true` threads the parameters of the system on to the next layer and therefore returns a `Tuple`, which could not be seeded at all. Worse, `composes_layerwise` said the chain decomposes, so `layerwise = :auto` committed to that path and then raised — [SNN #54](https://github.com/JuliaGNI/SymbolicNeuralNetworks.jl/issues/54).

SNN 0.7 widens the seam. `SymplecticEuler` now carries `SymbolicEnergy`'s `parameter_length` and `parameter_layout` — as a *trailing* type parameter, so every existing partial signature goes on matching with the layout left free — and implements the four methods that say how it meets the seam: `carried_variables`, `seam_value`, `state_expressions` and `seam_arguments`. What it carries is data and never a differentiation target: the sensitivity pairs with the state, and the flattened system parameters become one more argument of each generated kernel.

`ParametricLoss` declares its `loss_expression`, so the sweep can seed it. The construction guesses one by applying the loss in its four-argument form, and this loss takes five — without the declaration `loss_seed` would decline and fall back to the construction that cannot build. The declaration restates `_compute_loss`, which is what the functor itself uses.

## Results

```julia
SymbolicNeuralNetworks.SymbolicPullback(SymbolicNeuralNetwork(arch), FeedForwardLoss())
```

| case | before | now |
|---|---|---|
| `n_integrators = 1`, no system parameters | `MethodError` under `:auto` | builds layerwise; 1.4e-16 vs the monolithic path, 8.3e-17 vs `Zygote` |
| `n_integrators = 2` | expression of 10⁹ terms, never finished | 4 steps in ~1.5 s; 7.6e-17 vs `Zygote` |
| parametrized, `ParametricLoss`, 2 integrators | no construction builds | 2.2e-16 vs `Zygote` |

The parametrized row is the one worth dwelling on. The monolithic construction traces the chain from a plain vector, so every layer defaults its system parameters to `NullParameters` and the energy network's first `Dense` is then short of the components it reads; the build dies inside `Symbolics`. The layerwise construction is handed the pair `(state, system parameters)` and differentiates the map that was actually asked for. It is not the faster construction there — it is the only one.

## The tractability limit stays, and now says where to go

`_check_symbolic_pullback_is_tractable` keeps its `n_integrators > 1` refusal, because the two constructors in `src/pullbacks/symbolic_hnn_pullback.jl` still build one expression for the whole network by hand. It also guards `SymbolicPullback(arch)`, which had no guard at all and failed with an obscure `Symbolics` error.

That one cannot move onto the layerwise construction: `HNNLoss` evaluates the pre-built Hamiltonian vector field rather than the chain, so there is no prediction for a layerwise seed to start from. `SymbolicPullback(nn, ::ParametricLoss, system_params)` *can*, and doing so would also retire its concatenation of the system parameters onto the network input — that is #245's remaining half, left for a follow-up.

## Known limitation

The sweep runs the forward pass by *calling* each layer, so it inherits `SymplecticEuler`'s own single-sample restriction — the `@view(…)[:, 1]` in the layer and the `@assert size(qp, 2) == 1` in `concatenate_array_with_parameters`. A batch has to be taken sample by sample, as `apply_parametric` already does. The monolithic construction never calls the layer at run time and so does not have this restriction. Lifting it is layer work, independent of this PR.

## Testing

New `test/generalized_hamiltonian_neural_networks/layerwise_symbolic_pullback_test.jl`, registered in `runtests.jl`. All five GHNN test files pass, including the four that existed before — the struct gained fields and a type parameter, so those are the regression check.

## Blocked on

SymbolicNeuralNetworks 0.7 is not yet registered. Verified locally with that tree `Pkg.develop`ed in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,63 @@ breaking release).
 
 ## [Unreleased]
 
+### Added
+
+- **`SymplecticEuler` declares how it meets the layerwise pullback's seam**, which is what
+  [#245](https://github.com/JuliaGNI/GeometricMachineLearning.jl/issues/245) was waiting for.
+
+  `SymbolicNeuralNetworks` composes the pullback of a chain layer by layer, putting fresh symbolic
+  variables between two layers instead of inlining one layer's expression into the next. Its seam was a
+  plain vector, so it assumed every layer maps an array to an array; a `SymplecticEuler` built with
+  `return_parameters = true` threads the parameters of the system on to the next layer and returns a
+  `Tuple`, which could not be seeded at all
+  ([SNN #54](https://github.com/JuliaGNI/SymbolicNeuralNetworks.jl/issues/54)).
+  SymbolicNeuralNetworks 0.7 widened the seam; `SymplecticEuler` now carries `SymbolicEnergy`'s
+  `parameter_length` and `parameter_layout` and implements `carried_variables`, `seam_value`,
+  `state_expressions` and `seam_arguments`.
+
+  So
+
+  ```julia
+  SymbolicNeuralNetworks.SymbolicPullback(SymbolicNeuralNetwork(arch), FeedForwardLoss())
+  ```
+
+  builds for any `n_integrators` — four steps for two integrators in about 1.5 s, against a monolithic
+  expression of 10⁹ terms that never finished — and agrees with `Zygote` to machine precision.
+
+  For an architecture built with `parameters` it is the *only* construction that builds at all. The
+  monolithic one traces the chain from a plain vector, so every layer defaults its system parameters to
+  `NullParameters` and the energy network's first `Dense` is then short of the components it reads; the
+  build fails inside `Symbolics`. The layerwise construction is given the pair
+  `(state, system parameters)` and differentiates the map that was actually asked for.
+
+- **`ParametricLoss` declares its `loss_expression`.** The layerwise construction starts its sweep from
+  ∂L/∂ŷ and cannot guess one here — it applies a loss in its four-argument form, and this loss has five
+  — so it would have declined and fallen back to the monolithic construction, which for a
+  parameter-dependent architecture does not build. The declaration restates `_compute_loss`, which is
+  what the functor itself uses.
+
+### Changed
+
+- **`_check_symbolic_pullback_is_tractable` says where to go instead**, and now guards
+  `SymbolicPullback(arch)` as well as `SymbolicPullback(nn, ::ParametricLoss, system_params)`. Both
+  still build one expression for the whole network, so the `n_integrators > 1` limit is still real for
+  them — the first because `HNNLoss` evaluates the pre-built Hamiltonian vector field rather than the
+  chain, so there is no prediction for a layerwise seed to start from. `SymbolicPullback(arch)`
+  previously had no guard and failed with an obscure `Symbolics` error instead.
+
+  Migrating `SymbolicPullback(nn, ::ParametricLoss, system_params)` onto the upstream constructor —
+  which would also retire its concatenation of the system parameters onto the network input — is
+  [#245](https://github.com/JuliaGNI/GeometricMachineLearning.jl/issues/245)'s remaining half.
+
+### Known limitation
+
+- The layerwise sweep runs the forward pass by *calling* each layer, so it inherits `SymplecticEuler`'s
+  own single-sample restriction (`@view(…)[:, 1]`, and `@assert size(qp, 2) == 1` in
+  `concatenate_array_with_parameters`). A batch of more than one sample has to be taken sample by
+  sample, as `apply_parametric` already does. The monolithic construction does not have this
+  restriction, because it never calls the layer at run time.
+
 ### Removed (breaking)
 
 - **`RecurrentNeuralNetwork` and `LSTMNeuralNetwork`**, along with
--- a/src/architectures/generalized_hamiltonian_neural_network.jl
+++ b/src/architectures/generalized_hamiltonian_neural_network.jl
@@ -125,9 +125,15 @@ function build_gradient(se::SymbolicEnergy)
                                              inplace = false)
 end
 
-struct SymplecticEuler{M, N, FT<:Base.Callable, MT<:Chain, type, ReturnParameters} <: AbstractExplicitLayer{M, N}
+# `parameter_length` and `parameter_layout` are `SymbolicEnergy`'s, carried here because the layer is
+# what the layerwise `SymbolicPullback` asks about its seam -- see `carried_variables` below. `PT` is
+# the *last* type parameter so that every existing partial signature (`SymplecticEulerA{M, N, FT, AT,
+# false}` and the like) goes on matching, with the layout left free.
+struct SymplecticEuler{M, N, FT<:Base.Callable, MT<:Chain, type, ReturnParameters, PT} <: AbstractExplicitLayer{M, N}
     gradient_function::FT
     energy_model::MT
+    parameter_length::Int
+    parameter_layout::PT
 end
 
 function parameterlength(integrator::SymplecticEuler)
@@ -147,7 +153,9 @@ Changes ``q`` (based on the kinetic energy).
 function SymplecticEulerA(se::SymbolicKineticEnergy; return_parameters::Bool)
     gradient_function = build_gradient(se)
     c = Chain(se)
-    SymplecticEuler{se.dim, se.dim, typeof(gradient_function), typeof(c), :A, return_parameters}(gradient_function, c)
+    SymplecticEuler{se.dim, se.dim, typeof(gradient_function), typeof(c), :A, return_parameters,
+                    typeof(se.parameter_layout)}(gradient_function, c, se.parameter_length,
+                                                 se.parameter_layout)
 end
 
 """
@@ -156,7 +164,9 @@ Changes ``p`` (based on the potential energy).
 function SymplecticEulerB(se::SymbolicPotentialEnergy; return_parameters::Bool)
     gradient_function = build_gradient(se)
     c = Chain(se)
-    SymplecticEuler{se.dim, se.dim, typeof(gradient_function), typeof(c), :B, return_parameters}(gradient_function, c)
+    SymplecticEuler{se.dim, se.dim, typeof(gradient_function), typeof(c), :B, return_parameters,
+                    typeof(se.parameter_layout)}(gradient_function, c, se.parameter_length,
+                                                 se.parameter_layout)
 end
 
 # A network with no system parameters gets its input unchanged; without this the empty flat vector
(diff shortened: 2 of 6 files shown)


michakraus (commented):
**Verdict: request changes** — one blocker (the `[compat]` bound), one narrowing worth fixing before merge, and a few house-rule items. The substance of the change is right: I checked the four seam methods against the upstream contract, and `ParametricLoss`'s `loss_expression` against `_compute_loss`, and both hold.

Reviewed at `8e21ccf`, diffed as `origin/pghnn-rebase...origin/pr/251` (6 files: `CHANGELOG.md`, `src/architectures/generalized_hamiltonian_neural_network.jl`, `src/loss/losses.jl`, `src/pullbacks/symbolic_hnn_pullback.jl`, the new test file, `test/runtests.jl`). Everything that is only on `pghnn-rebase` is #207's and is not commented on.

**Which `SymbolicNeuralNetworks` I read the contract from.** The local checkout at `~/Research/Packages/SymbolicNeuralNetworks`, branch `main`, commit `faae4a2`, `version = "0.8.1"` — `src/derivatives/layerwise_pullback.jl`. The same four defaults are present verbatim at tag `v0.7.0` (`git show v0.7.0:src/derivatives/layerwise_pullback.jl` → `carried_variables` at 423, `seam_value` at 441, `state_expressions` at 459, `seam_arguments` at 478), and `git tag --contains 6700427` ("Let a layer say how it meets the seam") gives `v0.7.0` first. I did not read SNN PR #56 itself; it is merged and released.

---

## CI — a resolver failure, not a test failure, and not stale

All 13 jobs red, every one for the same reason. From the job log of *Julia 1.10 - ubuntu-latest*:

```
  [aed23131] + SymbolicNeuralNetworks v0.6.0
ERROR: LoadError: UndefVarError: `carried_variables` not defined
ERROR: LoadError: Failed to precompile GeometricMachineLearning …
```

The Documentation job says the same with the module named: ``UndefVarError: `carried_variables` not defined in `SymbolicNeuralNetworks` ``. No job ran a testset; all of them stop at precompilation.

This is **not** the expected "blocked on an unregistered dependency" any more, which is finding 1.

---

## Findings

### 1. blocker — `Project.toml:50` still says `SymbolicNeuralNetworks = "0.6"`, and 0.7 is registered now

This PR touches no `Project.toml`. The base branch needs none of the new names — `git grep -l 'carried_variables\|seam_value\|seam_arguments\|loss_expression\|state_expressions' origin/pghnn-rebase -- src test` returns nothing — so the requirement for SNN ≥ 0.7 is introduced here, and the bound that has to move with it is this PR's to move.

Two things changed since this was opened on 2026-08-24:

- **SNN 0.7 is no longer unregistered.** `General` now carries `0.7.0`, `0.7.1`, `0.8.0` and `0.8.1` (from `~/.julia/registries/General.tar.zst`, `S/SymbolicNeuralNetworks/Versions.toml`). The "Blocked on" section of the description is stale.
- So the bound is the only thing keeping CI red. The run on `8e21ccf` is the current head's run — there has been no newer push — and a re-run today fails identically, because `Pkg` still installs 0.6.0.

Worth deciding deliberately rather than reflexively writing `"0.7"`: **0.8.0 was a breaking release upstream** (`EquationSet` reinstated, `NeuralNetworkParameters` 0.3), and `main` here already sits at `SymbolicNeuralNetworks = "0.8"` with `AbstractNeuralNetworks = "0.8"`. None of the four seam methods added here touch what 0.8.0 broke, so `"0.7, 0.8"` looks reachable — but that is a claim to make with a green CI run behind it, not from reading.

### 2. bug (latent) — `_seam_state` covers `QPT`, while its caller accepts `QPT2`

`src/architectures/generalized_hamiltonian_neural_network.jl:303` takes `qp::QPTOAT2`, and `:312-313` handle it:

```julia
_seam_state(qp::AbstractArray) = qp
_seam_state(qp::QPT) = vcat(qp.q, qp.p)
```

`QPT` requires `Tuple{AT, AT}` — the *same* array type twice. `QPT2` is exactly the relaxation of that, and `src/utils.jl:159-161` says why it exists: *"A `Chain` that splits an input array into `q` and `p` produces views of different types, so the layers of a parameter-dependent network dispatch on this rather than on `QPT`."* That is not hypothetical here — `:204` returns `(q = @view(…)[:, 1], p = qp.p)`, a `SubArray` beside whatever `p` was.

Measured, with the two aliases copied verbatim out of `src/utils.jl`:

```
mixed_isa_QPT2 => true
mixed_isa_QPT  => false
same_isa_QPT   => true
mixed_seam     => MethodError: no method matching _seam_state(::@NamedTuple{q::SubArray{Float64, 1, Matrix{F…
same_seam      => qpt
```

The same gap is in the `Tuple` branch at `:292-296`, which calls `_seam_state(first(qp_params))`.

It is latent rather than live: the sweep reaches the seam through `batched`, which has methods for `AbstractVector`, `AbstractMatrix`, `AbstractArray` and `Tuple` and none for a `NamedTuple` (SNN `layerwise_pullback.jl:669-674`), so a `(q, p)` input fails upstream of this. Which leaves it either a narrowing that bites the first time a `QPT2` does arrive, or a `_seam_state(::QPT)` method nothing can reach. `QPT2` in that signature settles it either way, and matches how the rest of the file dispatches.

### 3. quality — history in comments, where the CHANGELOG already carries it

Comments describe the package as it is; how it got that way belongs in `CHANGELOG.md`, which in this PR says all of it already.

- `test/generalized_hamiltonian_neural_networks/layerwise_symbolic_pullback_test.jl:5-8` — "Until SymbolicNeuralNetworks 0.7 that seam was a plain vector … could not be seeded at all (SNN #54). It now says how it meets the seam, which is what GML #245 was waiting for."
- ibid. `:52` — "and used to be the ones that could not be seeded"
- ibid. `:66` — "`:auto` used to throw here"
- `src/pullbacks/symbolic_hnn_pullback.jl:54-55` — "`SymbolicNeuralNetworks` 0.6 added that construction … and 0.7 widened its seam"
- ibid. `:77` — "`SymbolicPullback(nn, loss, system_params)` predates the widened seam"

The `:54-55` one is the mildest, since a version bound is a fact about the dependency; "predates" and the three test-file ones are pure narration. The seam block in `generalized_hamiltonian_neural_network.jl:248-318` is clean by contrast — present tense throughout, and it reads better for it.

### 4. quality — two unused imports in the new test file

`layerwise_symbolic_pullback_test.jl:25` imports `layer_seed`; the file uses `checked_layer_seed` (`:54`) and never `layer_seed`. `:26` imports `layers` from `AbstractNeuralNetworks`; the only occurrences of that string in the file are prose (`:4`, `:13`, `:120`).

### 5. quality — the new test file is not formatted to `style = "sciml"`

```julia
format_file(f; overwrite = false, style = SciMLStyle())   # → false
```

Four hunks: the `using SymbolicNeuralNetworks:` continuation at `:24-25`, and the keyword-argument continuations at `:106`, `:130` and `:142`. The two source files the same check also reports are already non-conformant on `pghnn-rebase`, so those are not this branch's — but this file is entirely new, and `.githooks/pre-commit` runs `--check` on staged `.jl` and blocks.

### 6. correctness-risk (process) — "Resolves #245" auto-closes an issue this PR halves

The description opens with `Resolves #245`, which closes it on merge. Two paragraphs later the same description, and the new docstring at `src/pullbacks/symbolic_hnn_pullback.jl:77-80`, both say that migrating `SymbolicPullback(nn, ::ParametricLoss, system_params)` is *"#245's remaining half, left for a follow-up"*. #245's own "Suggested fix" is the general one — *"Do not trace the inlined chain. Compose the pullback layer by layer"* — and this lands it for the upstream constructor only.

GitHub closes the issue regardless of the sentence that follows. Either weaken the keyword (`Part of #245`) and close it by hand when the second half lands, or split the remainder into its own issue and point at that.

### 7. nit — `CHANGELOG.md:67`, `### Known limitation`

A heading that appears nowhere else in the file (`grep '^### '` gives `Added`, `Changed`, `Fixed`, `Removed (breaking)`, `Dependencies`, `Documentation`). It reads fine, but it is a sixth heading vocabulary for one entry; the paragraph sits just as well at the end of the `### Added` item it qualifies.

### 8. nit — two blank lines at `generalized_hamiltonian_neural_network.jl:319-320`

Before the `GeneralizedHamiltonianArchitecture` docstring. The formatter collapses it.

### 9. nit — two assertions in the new tests could be sharper

`layerwise_symbolic_pullback_test.jl:95-101` hand-rolls `try … catch error; error end` and then tests `raised isa ArgumentError`; `@test_throws ArgumentError` takes a message argument and does both, without shadowing `Base.error`. And `:122`'s `@test_throws Exception` passes on *any* exception — including an `UndefVarError` from a future typo in the line under test. The claim being made is narrower than that.

---

## Verified and holds

- **(c) `ParametricLoss`'s `loss_expression` does restate `_compute_loss`, and cannot drift.** The functor at `src/loss/losses.jl:278-281` is `_compute_loss(apply_parametric(model, input, system_parameters, params), output)`; the declaration at `:292` is `_compute_loss(ŷ, y)`. It is not a second expression that happens to agree — it is the same call to the same imported function (`src/GeometricMachineLearning.jl:90`, from `AbstractNeuralNetworks`), with the prediction substituted for the term that produces it. There is no second definition to diverge. The stated reason for needing it checks out too: SNN's guess is `loss(PassThroughLayer, ps, ŷ, y)` (`layerwise_pullback.jl:82-84`), four arguments, and `ParametricLoss` has only a five-argument method — so `checked_guess` catches the `MethodError` and declines.
- **(b) the trailing type parameter claim holds.** Every partial signature in the file leaves `PT` free: the two aliases at `:147-148` (five parameters), the four functors at `:202`, `:207`, `:212`, `:217` through those aliases, the two array functors at `:230`, `:238` (six), and `state_expressions` at `:286` (six). Both constructors at `:156-158` and `:167-169` pass all seven, and the argument order `(gradient_function, c, se.parameter_length, se.parameter_layout)` matches the field order at `:133-136`. Nothing constructs or destructures a `SymplecticEuler` outside this file — `grep -rn SymplecticEuler src test docs ext` gives only `SymplecticEulerA/B(se; return_parameters = …)` in `forced_generalized_hamiltonian_neural_network.jl:29,32`, the unrelated `SymplecticEulerIntegrator*` training methods, and prose.
- **(d) the widened guard still refuses what it refused.** `_n_symplectic_integrators` (`symbolic_hnn_pullback.jl:32-33`) and the `n > 1` condition (`:84`) are untouched; only the message changed. The new call site is `:17`, the first line of `SymbolicPullback(arch)`, ahead of `SymbolicNeuralNetwork(arch)` — so it is reached, and for a `GeneralizedHamiltonianArchitecture` it reads `arch.n_integrators`. The pre-existing call at `:114` is unchanged. `test:95-101` exercises the newly guarded path.
- **(a) the four methods match the contract as SNN states it.** `carried_variables` returns `()` for a parameter-free system, which is what the docstring asks for in as many words, and otherwise a 1-tuple of `:π` variables — distinct from the `:x` and `:λ` that `layer_seed` uses, which `_assert_distinct_seam_variables` (`layerwise_pullback.jl:352-358`) requires and whose violation would be a silently wrong gradient. `seam_value`'s two arities line up with the default `seam_value(layer, sx, carried...)` and are strictly more specific than it. `state_expressions` is declared only for `return_parameters = true`; the `false` case correctly falls through to SNN's default `scalar_expressions(y)`, since that functor returns the array alone. `seam_arguments` returns `(state, carried)` in the order `carried_variables` declared, broadcast to the state's column count — the "same rank and batch size as the state" constraint at `layerwise_pullback.jl:386-389`. The `unflatten`/`_flatten_system_parameters` round trip between build time and run time is the pairing `concatenate_array_with_parameters` uses, and the one `SymbolicPullback(nn, ::ParametricLoss, …)` already uses at `symbolic_hnn_pullback.jl:119-123`.
- The `ArgumentError` in `seam_arguments(…, ::QPTOAT2)` is not error handling for a case that cannot occur. `adjoint_step` calls `seam_arguments` *after* the forward pass (`layerwise_pullback.jl:715-719`), so it is reached exactly when the forward pass survived an input it should not have — and the alternative there is a carried datum of the wrong length that nothing checks. Failing loudly is right.
- **(e) the new test file is registered and asserts on values.** `test/runtests.jl:52-54`, inside `@safetestset`. It compares against `Zygote` in four places and against the monolithic construction in one, all at `< 1e-14` — a tight bound for a gradient of this size, not a loosened one, and the PR's own reported agreements are 1e-16 to 1e-17. It also pins structure that would otherwise regress silently: `length(symbolic_steps(snn)) == 2n`, `composes_layerwise(snn)`, `construction(pullback) == :LayerwiseGradientFunction` (so a silent fallback to the monolithic path fails the test rather than passing it), and `isnothing(first(steps).dλ)`.
- **No piracy in what is added.** All four seam methods and `loss_expression` are SNN functions dispatched on GML types. (`Jacobian` at `:74` and the `SymbolicPullback` functors at `symbolic_hnn_pullback.jl:141` are piracy, are already marked `TODO`, and are not this diff's.)
- **NFC.** `nfc.jl` exits 0 on all six changed files — `π`, `λ`, `ŷ`, `⁹` and the rest are normalised.
- **CHANGELOG.** Present, substantial, under `## [Unreleased]`, and it says what the change means for a caller rather than which files moved.

## Not checked

- **The test suite did not run, and that is neither a pass nor a fail.** It cannot resolve: `[compat]` pins SNN 0.6 while the code needs ≥ 0.7, and editing `Project.toml` or resolving in the package environment was out of bounds for this review. Finding 1 has to be fixed before anybody — CI included — can say whether these tests pass. The description's "All five GHNN test files pass" is therefore unverified here, not disputed.
- **The performance figures are not re-measured** — the 10⁹ terms, the ~1.5 s for four steps, and the 1.4e-16 / 8.3e-17 / 7.6e-17 / 2.2e-16 agreements. Same reason.
- **Aqua, ExplicitImports, `detect_ambiguities` and `@code_warntype` were not run**, for the same reason: the package does not load against the registered SNN. `seam_value`'s two arities against the upstream `seam_value(layer, sx, carried...)` default are the one place I would want `detect_ambiguities` to confirm what I read off the specificity rules.

PR #254, opened 2026-08-25
Name the parameter union, and correct a comment 0.5.0 falsified

michakraus:
Two changes, one mechanical and one a correction that was overdue by a release.

## `ParameterSet`

Fifteen signatures here spelled `Union{NamedTuple, NetworkParameters}` out inline. That union is **`NeuralNetworkParameters.ParameterSet`** now, added upstream in 0.2.2: the same union, in the package that owns the type. `SymbolicNeuralNetworks` had named it `EquationSet`, `AbstractNeuralNetworks` spelled it out eight times, `GeometricOptimizers` sixteen — each of those with a comment explaining why its own narrower alias, `ParameterContainer{T}`, would not fit. One name now.

The sites are across `loss/losses.jl`, `loss/hnn_loss.jl`, `optimizers/optimizer.jl`, `pullbacks/zygote_pullback.jl` and `data_loader/optimize.jl`. Definitionally identical, so nothing about dispatch changes — which makes the full suite passing the property worth checking rather than a formality: a test that *moved* would have meant the substitution was wrong, not that a finding had been made. None did.

## A comment that `GeometricOptimizers` 0.5.0 falsified

Found in the review of [GeometricOptimizers #68](https://github.com/JuliaGNI/GeometricOptimizers.jl/pull/68). The note on `_make_optimizer_cache`'s branch ordering read:

> Today that ordering is invisible: `NetworkParameters` is not one of the types `GeometricOptimizers.OptimizerSolution` unions, so `_use_go_cache` is false at the root anyway…

`GeometricOptimizers` 0.5.0 **made the container a member of that union**. So `_use_go_cache` is true at the *root* now, and asking the structural question first is load-bearing rather than invisible: without the `x isa NetworkParameters` test coming ahead of it, a whole network would get **one** cache instead of one per layer — silently, with `_leaf_optim_step!` handed the entire tree and `_GMLGradient` having no method for it.

**The code was already right.** The ordering went in ahead of the change it anticipated, which is the whole point of having written it that way; only the comment needed correcting. That is a good outcome for the reasoning, and a reminder that a comment asserting "this is currently invisible" has a shelf life.

The note now also records what the better end state is, now that `GeometricOptimizers` 0.6.0 takes a whole container as one solution: **one cache for the network**, which means giving `_GMLGradient` a `NetworkParameters` method and dropping `_tree_optim_step!`. That is a change of behaviour — one `GlobalSection` tree and one `Q` across every layer instead of one per layer — so it wants its own release rather than being folded in here.

## Blocked on

`NeuralNetworkParameters = "0.2.2"`, so [JuliaGNI/NeuralNetworkParameters.jl#14](https://github.com/JuliaGNI/NeuralNetworkParameters.jl/pull/14) has to land and be registered first. Verified locally against that branch and against `AbstractNeuralNetworks` [#37](https://github.com/JuliaGNI/AbstractNeuralNetworks.jl/pull/37).


michakraus:
Reviewed as part of the `NeuralNetworkParameters` 0.2.2 chain. The fifteen inline unions and the comment correction are both right; three things to add, and one of them is a claim in this PR that I checked and can confirm.

## Item 5 confirmed: the comment was wrong for a release, and the code was not

`src/optimizers/optimizer.jl:52-68` said the branch ordering in `_make_optimizer_cache` was invisible because `NetworkParameters` is not one of the types `GeometricOptimizers.OptimizerSolution` unions. **GO 0.5.0 made it one**, so `_use_go_cache` is true at the *root* now and the ordering is load-bearing: without it a whole network would get one cache instead of one per layer, silently, with `_leaf_optim_step!` handed the entire tree. The ordering had been put in ahead of the change it anticipated, so only the comment needed correcting. Verified against GO 0.5.0's `optimizer_solution.jl`.

The same widening reaches `_tree_optim_step!`'s `λY isa ParameterSet` test at `:298`, correctly.

## `_tree_optim_step!` should stay hand-written, and now says why

`NeuralNetworkParameters`' `PLAN.md` listed it as a walk `foreachparameters` could replace. It cannot, for two reasons that are both load-bearing and are now recorded at `:270-292`: the recursion is keyed on the **cache** tree and stops where the cache stops, so a layer's `NamedTuple` reaches `_leaf_optim_step!` whole — which is what one `GeometricOptimizers` cache is for — and `λY` is **broadcast, not zipped**, one `GlobalSection` standing in for a whole subtree. That is a different walk, not a copy of the generic one.

While checking that, three more items `PLAN.md` listed as remaining here turned out to be **already done**: `src/map_to_cpu.jl` is one `mapstorage` call, `apply_toNT` and `_eltype` are gone, and the five `changebackend` methods moved to `GeometricOptimizers`' `ext/AbstractNeuralNetworksExt.jl` in 0.5.0 — which closes that document's D3 and the second half of its D8, and `GlobalSection(::NetworkParameters)` closes D11. `ext/HDF5Ext.jl` is **86 lines**, not the 119 recorded. All corrected upstream.

## Two bumps this PR needs to ship

- **`version` was left at 0.6.0**, so no release would have carried the rename or the tightened `NeuralNetworkParameters = "0.2.2"` bound. **0.6.1**.
- **`GeometricOptimizers = "0.6"`.** 0.6.0 is what takes a whole `NetworkParameters` through the optimizer and it is breaking — four `outer!`/`_mul!` methods, one `update!(::BFGSState, …)`, `add!` on a parameter set, and two `l2norm` methods that moved upstream to `GeometricBase`. **None had a caller here**; the `add!` arm was deleted on this side in 0.7 for the same reason. So the bump is a compat entry and no code.
- **Julia 1.11** is the floor, in step with the wave. Nothing here was accommodating 1.10: the two places naming it record where a measurement was taken and how a defect was found, and both stay as the history they are.

## Verified

**Full suite green on Julia 1.11.9 against the whole dev'd chain — 58 testsets**, including the HDF5 round trip for the GML special array types, `changebackend`, the optimizer convergence tests and every docstring example.

## One commit is held back

The version/compat/floor commit is **not pushed**. This repository's pre-push hook runs `Pkg.test()` against the registry, and it cannot resolve `GeometricOptimizers = "0.6"` until GO 0.6.0 is registered — the same blocker that makes every job on this PR red today (`Unsatisfiable requirements detected for NeuralNetworkParameters`). It will go up as soon as the chain registers rather than being forced past the hook.

**Merge order**: NNP #14 → register 0.2.2 → [GeometricBase #15](https://github.com/JuliaGNI/GeometricBase.jl/pull/15) → [SimpleSolvers #185](https://github.com/JuliaGNI/SimpleSolvers.jl/pull/185) → GO #68 → GO #69 → register GO 0.6.0 → ANN #37 → SNN #60 → **this** → GMLDatasets #10 → NonlinearIntegrators.

michakraus:
Pushed one more commit: **`Set version to 0.6.1, take Julia 1.11 and GeometricOptimizers 0.6`**.

It was held back until `GeometricOptimizers` 0.6.0 was registered, because until then no runner could
resolve the compat entry it adds — every job on this PR would have failed for that reason alone, which
is exactly the noise the rest of this wave has been generating. 0.6.0 is in the General registry now,
so the matrix here is meaningful for the first time.

Three things in it:

- **Version 0.6.1.** The commit before it replaced fifteen inline `Union{NamedTuple, NetworkParameters}`
  with `NeuralNetworkParameters.ParameterSet` and corrected the `_make_optimizer_cache` comment, and
  left the version at 0.6.0 — so no release carried either. This one does.
- **`GeometricOptimizers = "0.6"`.** 0.6.0 is breaking: four `outer!`/`_mul!` methods, one
  `update!(::BFGSState, ...)` and `add!` on a parameter set are gone, and two `l2norm` methods moved
  upstream to `GeometricBase`. None of them had a caller here — the `add!` arm was deleted on this side
  in 0.7 for the same reason — so this is a compat entry and no code.
- **Julia 1.11 is the floor**, matching the rest of the stack. Nothing here was accommodating 1.10: the
  two places that name it, `test/training_parameters.jl` and
  `test/optimizers/optimizer_convergence_tests/psd_optim.jl`, record where a measurement was taken and
  how a defect was found. Both stay as the history they are.

The CI matrix drops 1.10 accordingly, so the thirteen red checks above — all of which are either the
1.10 row or an unresolvable `NeuralNetworkParameters` — should not recur.

PR #261, opened 2026-09-14
Regenerate the inverse kernels with CSE: 25.5 MB to 27 KB

michakraus:
`src/kernels/inverses/inverse_5x5.jl` was **25 495 007 bytes and 91 625 lines** — 96 % of the
package's `src/`. `fatou lint` did not terminate on it, nor on `src/GeometricMachineLearning.jl`,
which reaches it through the include chain; the shared `.githooks/pre-commit` runs `fatou lint` on
staged files with no timeout, so staging either file hung the commit.

`Symbolics.build_function` defaults `cse` to `false`. All four kernels are regenerated with
`cse = true`.

| | before | after | |
|:--|--:|--:|--:|
| `inverse_2x2.jl` | 1 875 B / 53 lines | 1 706 B / 59 lines | ×1.1 |
| `inverse_3x3.jl` | 12 450 B / 221 lines | 3 919 B / 98 lines | ×3.2 |
| `inverse_4x4.jl` | 76 763 B / 1 175 lines | 10 228 B / 209 lines | ×7.5 |
| **`inverse_5x5.jl`** | **25 495 007 B / 91 625 lines** | **26 657 B / 487 lines** | **×957** |
| total | 25 586 095 B | 42 510 B | ×602 |

`fatou lint` on the regenerated 5×5: **45 ms, clean**. On the whole `src/` tree: **0.1 s**.

## The size was whitespace, not mathematics

**24 941 056 of the old file's 25 495 007 bytes — 97.8 % — were leading indentation**, at up to 512
columns. Stripped of whitespace the file is 450 370 bytes.

Without CSE each entry of the inverse is one deeply nested expression, and JuliaFormatter indents
every level of that nesting at a cost quadratic in the depth. That is why the 4×4, from the same
generator with the same defaults, is only 77 KB: the depth had not yet become the dominant term.

## The 5×5 had no test coverage

`test55_inverse()` and `test55_inverse_pullback()` in `test/kernels/tensor_inverse.jl`, and
`test_tensor_cayley5` in `test/kernels/tensor_cayley.jl`, were defined and never called. All three
are enabled and pass.

## One assertion changed, and why it is not a tolerance being widened

`test/attention_layer/attention_setup.jl` asserted `det₁ ≈ det₂` and then `det₂ ≈ det₃` — two
independently computed approximations against each other, so their errors add. At N = 3 in
`Float16` each is about 2 % out from the exact `det₁`, inside the 3.125 % `Float16` tolerance on its
own; the second assertion held only while the two errors pointed the same way. The regenerated 3×3
changes the rounding of `det₂` alone and they then point opposite ways.

It now asserts `det₁ ≈ det₂` and `det₁ ≈ det₃`. That is what volume preservation means, and it does
not double the error budget.

**The 2 % is not new.** `det₃`'s deviation is bit-identical before and after. And accuracy did not
regress: against a `Float64` reference over 2000 random slices per size, mean relative error with
CSE is the same or **lower** at every size and precision but one — `Float16` 5.6e-4 vs 6.0e-4 (3×3),
9.1e-4 vs 9.3e-4 (4×4), 1.9e-3 vs 3.0e-3 (5×5); `Float32` 2.1e-7 vs 4.6e-7 (5×5). The 2×2 is
bit-identical.

## Also here

- `legacy/codegen/matrixinverse.jl` emits the complete kernel file — the slice index, the Cartesian
  output index, the `tensor_inverseN` wrappers and the `rrule` — rather than a `build_function` body
  that a human then wraps. Running it and then JuliaFormatter with `style = "sciml"` reproduces all
  four committed files byte for byte.
- The `invNN_kernel!` signatures are **widened**. The 2×2, 3×3 and 4×4 were
  `(ˍ₋out::AT, A::AT) where {T, AT <: AbstractArray{T, 3}}`, so `tensor_inverseN!(out, A)` with a
  plain `Array` output and a `SubArray` input was a `MethodError` — the same defect class as the
  `::AT` cotangent signatures already fixed in this release. They now match the 5×5, which never
  carried the annotation.
- `simplify.(B)` is inert under Symbolics 7.39.2: the generated string is byte-identical with and
  without it, at every n from 2 to 5. The `if n ≤ 6` guard is left in place as the record that
  `simplify` is impractical above 6.

`Pkg.test()` passes on this branch, exit 0, zero failures.

## Pre-PR verification

**Fixed**
- `legacy/codegen/matrixinverse.jl:1` — the header comment promised the files in
  `src/kernels/inverses/`, but the script writes unformatted output into the working directory. The
  comment now names the JuliaFormatter `style = "sciml"` pass that the reproducibility claim depends
  on.

**Unresolved** — none.

**Pre-existing, adjacent** (reported, not fixed)
- `test/kernels/tensor_inverse.jl:24-29` and four siblings — the pullback loops index `k`, not the
  loop variable `i`, so each checks the last slice `k` times. Owned by branch
  `fix-pullback-slice-index`. This branch enables `test55_inverse_pullback()`, so it propagates the
  defect into newly-run code; it does not create it. **bug**
- `test/kernels/tensor_cayley.jl:16,22,28,34` — the `T::Type` parameter is unused and
  `rand(n, n, d)` is always `Float64`, so `check_all(Float32)` repeats the `Float64` run. The
  `Float32` path of the newly enabled 5×5 Cayley stays untested. **bug**
- `test/runtests.jl:47` and `:148` — `kernels/tensor_inverse.jl` is included twice. Owned by branch
  `remove-duplicate-tensor-inverse-include`. **quality**
- `test/attention_layer/attention_setup.jl:28-36` — `check_all` covers N = 10, 2, 3, 4 but not 5, so
  `tensor_cayley5` is not reached through `VolumePreservingAttention`. **nit**
- `src/utils.jl:61,67,163` — three genuine piracies on Base functions over Base types. Outside the
  diff. **correctness-risk**

**Checked and clean**
- Correctness: running the generator and then JuliaFormatter reproduces all four committed kernels
  **byte for byte**. That is the strongest available check on the 91 000 deleted lines.
- Correctness: the assertion change is required. At N = 3 in `Float16`, `det₂ ≈ det₃` is false on
  this tree — deviations 1.87 % and 1.92 % against a 3.125 % tolerance, 3.86 % apart. Both new
  assertions hold.
- Correctness: the `invNN_kernel!` signatures were widened, so no caller can break. The wrappers and
  `rrule`s are byte-identical to the base.
- Type piracy: `Aqua.Piracy.hunt` gives 13 hits, none with a definition site in the diff. Aqua is
  not a dependency of this package, so this is a first run with no baseline.
- Type instability: `code_typed` on all four generated CPU kernels gives 0 `Any`-typed call
  statements (5×5: 0 of 31 298). `code_warntype` on `tensor_inverse5!` is clean in `Float64`,
  `Float32` and `Float16`.
- Allocations: fresh process, `--check-bounds=auto`, minimum of five after warm-up —
  `tensor_inverse2!` through `tensor_inverse5!` all allocate exactly 80 bytes. Constant in kernel
  size, so it is launch overhead and the regeneration did not move it.
- `fatou lint`: all eight changed files clean in 0.2 s. JuliaFormatter: all already formatted.
  NFC: passes.
- Comments: present tense and self-contained. The 98 % leading-whitespace claim measures 97.83 %;
  every CHANGELOG byte and line figure reproduces exactly.
- History: no `previously`/`used to`/`no longer`/version/date in any added line outside
  `CHANGELOG.md`.

**Not checked**
- GPU backends — only `CPU()` is available in this environment. The kernels are `@kernel`, so the
  CUDA/Metal specialisations are unverified.
- Doctests — the diff touches no `docs/` file.
- ExplicitImports and `stale_deps` — the diff changes no import or dependency.

Works `Tasks/Shrink the generated 5x5 inverse kernel.md`. Whether to rewrite history to drop the
four 25 MB revisions from `.git` is **not** decided here.


michakraus (commented):
## Review: approve, with two corrections to the accuracy paragraph — and the PR undersells itself

Reviewed at `4728f49f` against `58963c0e` in a fresh detached worktree, with a second worktree at the
base for A/B. Everything below is measured in this session. Where I reproduce one of your figures I
say so.

91 625 deleted lines cannot be reviewed by reading. Two things stand in for that, and both hold: the
generator reproduces all four files byte for byte, and the kernels are correct against an independent
reference.

### The headline is not the file size

The PR measures allocations and finds them unchanged. It never measures runtime. It should, because
this is the largest effect in the change:

| `tensor_inverse5!`, `Float64` | base | this branch | |
|:--|--:|--:|--:|
| per slice, min of 200 after warm-up | 969.7 ns | 31.7 ns | **×30.6** |
| first call in a fresh process | 10.785 s | 0.350 s | **×30.8** |

Cold process each time, `--check-bounds=auto`, 4096 slices. Repeat run: 969.8 ns / 32.5 ns and
10.846 s / 0.292 s. The 4×4 goes 55.8 → 11.1 ns and the 3×3 12.2 → 3.1 ns; at the 2×2 both sit near
the noise floor (2.4 → 1.1 ns) and the output is bit-identical, so I would not claim that one.

**Ten seconds of compile time per process, for one kernel.** That is worth a line in the CHANGELOG on
its own, and it is a plausible reason nobody ever un-commented `test55_inverse()`.

### Why — the mechanism, which is worth stating

Counting one output entry of the base 5×5: 696 `getindex` and 624 multiplications. That is the fully
expanded Leibniz determinant — 120 monomials × 5 factors = 600 — plus a 4×4 cofactor numerator of
24 × 4 = 96. Written out again for each of the 25 entries: about 15 600 multiplications per slice,
and 25 × 17 983 = 449 575 characters of expression text. That is where the 450 370 whitespace-free
bytes you report come from — the remaining 795 are the wrappers and the `rrule`.

The regenerated 5×5 is 353 `getindex` and 419 multiplications **for the whole kernel**. The
determinant is a nested cofactor expansion built once as `t102`, and all 25 entries divide by it
(exactly n² divisions per file, at every size).

That one change drives all three results: the size, the speed above, and the accuracy below — a
nested cofactor expansion cancels far less than a sum of 120 degree-5 monomials. It also explains why
the 2×2 comes out bit-identical: with two monomials there is nothing to cancel.

### Verified

**Reproducibility.** In an isolated environment — Symbolics 7.39.2, JuliaFormatter 2.14.0,
`style = "sciml"` — the generator plus the formatter reproduces all four committed files with
matching MD5s and matching sizes. `cmp` exits 0 on all four. This is the strongest available check on
the deleted lines and it passes.

**The regex rewriting did not corrupt anything.** All n² output cells assigned exactly once
(4 / 9 / 16 / 25 distinct `ˍ₋out[i, j, k]`); no surviving `##cse#` names; no stray linear
`ˍ₋out[N]`; no un-sliced `(getindex)(A, i, j)`; every temporary assigned once and defined before
first use, in all four files.

**Numerical correctness.** Round trip `‖A·A⁻¹ − I‖ < 1e-10` at all four sizes. Forward error against
`inv` on a `Float64` copy of the same entries, 4000 slices per size, bucketed by condition number so
the draw does not contaminate the verdict. A/B over identical draws, mean relative error, this branch
against base:

| | cond < 10 | cond 10–100 |
|:--|:--|:--|
| `Float64` 5×5 | 3.19e-16 vs 5.34e-16 | 8.27e-16 vs 1.84e-15 |
| `Float32` 5×5 | 1.45e-07 vs 2.74e-07 | 4.03e-07 vs 9.51e-07 |
| `Float16` 5×5 | 1.19e-03 vs 2.19e-03 | 3.37e-03 vs 7.92e-03 |

The 2×2 is bit-identical in all three precisions — equal output hashes over 4000 slices, as claimed.

**The assertion change is required, not a tolerance being relaxed.** Replaying the file's exact draw
sequence: at N = 3 in `Float16`, `det₂` is 1.87 % from the exact determinant and `det₃` is 1.92 %,
they point opposite ways, and the tolerance is 3.125 %. `det₂ ≈ det₃` fails. Both replacement
assertions hold, and they hold at all twelve (N, T) combinations, so the change costs no coverage
elsewhere. Your figures reproduce exactly.

**The lint hang is real.** `fatou lint` on the old 25 MB file does not terminate within 90 s; the
`timeout` cap kills it. The eight changed files lint clean in 45 ms.

**The `MethodError` claim is real.** On the base, `tensor_inverseN!(::Array, ::SubArray)` raises
`MethodError` at 2×2, 3×3 and 4×4 and works at 5×5 — exactly as described. It works at all four sizes
here.

**Every CHANGELOG figure reproduces.** 25 495 007 B / 91 625 lines; 26 657 B / 487 lines; 42 510
against 25 586 095 total; 24 941 056 bytes of leading indentation = 97.83 %; 450 370 bytes stripped
of whitespace; 512 columns of maximum indent.

Formatting: all eight files already formatted. NFC: clean — and I confirmed the checker flags a
deliberately decomposed file, so that is a real negative and not a silent pass. No `docs/` file
references these kernels.

**One caveat on the coverage number**, before anyone quotes it. Codecov reports 65.58 % → 72.75 %,
+7.16 points, but most of that is the denominator moving: CSE turns one unmeasurable expression into
hundreds of countable statements, so the line count rises by 628 and 669 of the new lines are hits.
The honest signal is **misses falling by 41** — that is the 5×5 going from untested to tested.

### Please fix: the accuracy paragraph names an exception it never identifies

> "mean relative error is the same or lower than before at every size and precision but one"

The list that follows gives four cases where it is **lower**. The "one" is never named. Either name it
or drop the clause.

From my A/B the exceptions sit in the ill-conditioned tail (cond ≥ 100) and are small: `Float64` 3×3
2.37e-14 against 1.72e-14, `Float32` 3×3 7.92e-06 against 6.46e-06, `Float32` 4×4 4.65e-06 against
4.05e-06.

### Please add one sentence: `Float16` 5×5 overflows where the base did not

Two slices in 4000 `rand(Float16, 5, 5, ·)` draws come back non-finite on this branch and finite on
the base, at condition numbers 4.45e3 and 7.84e3. It is **not** a zero denominator — the `Float16`
determinant is a normal number, 2.49e-4. The true inverse entries are O(10⁴), near the `Float16`
ceiling of 65504, and the changed rounding tips them over.

This does not block the merge, for four reasons, and I would rather the CHANGELOG said so than leave
it to be rediscovered:

- It is symmetric. The base has the same failure at 3×3 (cond 2.98e4), which this branch **fixes**.
  The 2×2 failure (cond 1.54e4, determinant exactly zero in `Float16`) is identical on both sides.
- The lowest failing condition number is 4.45e3. `Float16` carries about 3.3 decimal digits; the
  answer is meaningless there whichever expression computes it.
- **The package cannot reach that regime.** `tensor_inverse5` is called only from
  `tensor_cayley5(S) = (I − S)·inv(I + S)` with `S` skew-symmetric, and `cond(I + S)` over 4000
  `Float16` draws is **[1.08, 2.18]** — three orders of magnitude away. `tensor_cayley5` gave 0
  non-finite entries in 4000 draws.
- In the regime the package does use, this branch is uniformly more accurate.

### Nits

**The `if n ≤ 6` guard can no longer be false**, now that the loop is `for n in 2:5`. And
`simplify.(B)` is confirmed inert: the generated output is byte-identical with the block and without
it at every size, and the simplify-free variant still reproduces the four committed files. It costs
0.66 s at n = 5, about 0.9 s of script time. An inert call kept as documentation reads as a live
step; a comment carries the same record. Your call — the generator runs by hand.

**The history rewrite is not worth scheduling.** You leave this open. Measured: the 25 495 007-byte
blob packs to 464 650 bytes, and all five distinct `inverse_5x5.jl` blobs together are 0.5 MB of a
420 MB pack. The expensive part was always the 25 MB **checkout**, which this branch already fixes. I
would close the question rather than carry it.

Relatedly, "Every clone, CI checkout and `Pkg.add` also stops moving a 25 MB file" is right for the
checkout and overstated for the clone, which transferred about 0.46 MB.

### Pre-existing defects — confirmed, correctly attributed, not this branch's to fix

All three are reported in the PR body and all three have open siblings. I confirm each, and note that
enabling the 5×5 tests is what makes them matter:

- `test/kernels/tensor_inverse.jl` pullback loops index `k`, not `i`, so the newly enabled
  `test55_inverse_pullback()` gives one slice of real coverage rather than ten. → #260
- `test/kernels/tensor_cayley.jl` ignores its `T` parameter — `rand(n, n, d)` is always `Float64` —
  so the newly enabled 5×5 Cayley has no `Float32` coverage. → #263
- `test/runtests.jl:47` and `:148` include the file twice, both inside `@safetestset`, so the newly
  enabled 5×5 tests run twice. → #262

The ordering is right: this branch first, then the three that sharpen what it turned on.

### CI

All nine `CI.yml` matrix jobs pass — `Julia 1`, `min`, `pre` and `nightly` across ubuntu, macOS and
Windows — plus `codecov/patch` and `codecov/project`.

`Documentation` fails, and it is **known-red at the base**. I compared the two logs: the error sets
are identical, 66 errors each, with no difference in either direction. One `@example` failure in
`docs/src/tutorials/optimizer_comparison.md:3-36` (a `MethodError` on
`Optimizer(::GradientMethod, ::@NamedTuple{xy::Matrix{Float64}}; step_size::Float64)`), two
unresolved `@ref`s, and 63 missing images. Nothing in the log mentions the kernels, and this branch
touches no `docs/` file. Not a regression.

(Worth knowing: `gh pr checks` reported `Julia min - windows-latest` as pending for a long time after
the job had actually concluded `success`. `gh run view` was right and `gh pr checks` was stale.)

---

PR #262, opened 2026-09-14
Include the tensor-inverse tests once, not twice

michakraus:
`test/runtests.jl` included `test/kernels/tensor_inverse.jl` from two `@safetestset` blocks, and
neither had a GPU guard, a Julia-version guard, or any other condition:

- `"Custom inverse for 2x2, 3x3, 4x4, 5x5 matrices"`
- `"Test parallel inverses"`

So a full `Pkg.test()` evaluated the same 110 assertions 220 times for no added coverage. This
removes the first block.

## Which block stays, and why

`"Test parallel inverses"`. The choice comes from the history, not from taste:

| commit | date | what it did |
|:--|:--|:--|
| `be7dc9ba` | 2024-04-08 | added `test/kernels/tensor_inverse.jl` and `tensor_cayley.jl` |
| `4b4084ad` | 2024-04-08 | added `"Test parallel inverses"` **and** `"Test parallel Cayley"` together |
| `7bf59f6f` | 2024-04-12 | "Added new DataLoader tests." — also added `"Custom inverse for 2x2, …"` |

`"Test parallel inverses"` entered with the two kernel test files themselves, alongside its
`"Test parallel Cayley"` sibling, and still sits next to it. The duplicate arrived four days later
inside an unrelated data-loader commit — the same commit that added a second `include` of
`test/arrays/triangular.jl` next to one that already existed. One slip, twice in one diff.

**The two halves of `runtests.jl` were never a CPU/GPU split.** The second half is simply where the
file grew by appending, and the `@info` banners in it were added much later (`8dc9b831`,
2026-08-12) as progress markers only.

`test/kernels/tensor_cayley.jl` is included exactly once and is untouched. After this change no
`include` in `test/runtests.jl` is repeated.

## Verification

- `"Test parallel inverses"` (110 tests) and `"Test parallel Cayley"` (60 tests) both pass, run by
  line range with `testrunner --project=. test/runtests.jl -f=144:146` on Julia 1.13,
  `--check-bounds=auto`.
- `test/kernels/tensor_inverse.jl` is reached exactly once, from `test/runtests.jl:145`.
- All 59 `include` calls in `test/runtests.jl` are distinct, and every included path exists.
- No reference to the removed testset name survives outside `CHANGELOG.md`.
- JuliaFormatter `--check`, `fatou lint` and the NFC check are clean; the file parses; the package
  loads. All four pre-commit stages passed.
- Every factual claim in the CHANGELOG bullet was re-derived independently: the 110 assertions
  counted from the file, and the commit archaeology confirmed with `git log -S` and `git show`.

Type piracy, inference and allocations were not checked, and doctests were not run: no `src/` or
`docs/` file changed and no callable symbol was touched.

## Pre-existing, not fixed here

**40 test files under `test/` are included from nowhere** — among them all of
`test/performance_tests/`, `test/orthogonalization_procedures/` (5 files), `test/cuda/`, and
`test/train!/test_training.jl`. This branch orphaned none of them, and each needs its own decision
(wire up, or delete). It is the same class of harness rot as the defect fixed here.

## Coordination

PR #260 (`fix-pullback-slice-index`) also edits `CHANGELOG.md` and `test/kernels/tensor_inverse.jl`.
This branch touches `test/runtests.jl`, so the only overlap is the changelog. This bullet is placed
at the **end** of the `### Fixed` list, while #260 inserts at the top, so the two should merge
without a conflict in either order.


michakraus (commented):
**Verdict: approve.** Rebased onto `main` at `e88aafab` (the #261 merge), reviewed at `2aa3359b`.

## Rebase

Clean, no conflicts. The branch is one commit. The CHANGELOG bullet sits at the end of
`### Fixed`, which is where it was placed to avoid a clash with #260, and that held.

## One finding, fixed on the branch

The CHANGELOG and the commit message both said the duplicate ran **"the same 110 assertions 220
times"**. That figure was measured before #261. #261 enabled `test55_inverse()` and
`test55_inverse_pullback()`, which add 30 assertions, so the rebase invalidated the number.

Measured on the rebased tree — `Pkg.test(; test_args=["--filter=Test parallel inverses"])`,
Julia 1.13:

```
Test parallel inverses                                                           |  140    140  0.7s
Test parallel Cayley                                                             |   80     80  0.1s
```

Both pass. The commit now says 140 and 280.

## Checked and clean

- `test/kernels/tensor_inverse.jl` is reached exactly once, from `test/runtests.jl:145`.
- All 59 `include` calls in `test/runtests.jl` are distinct, and every included path exists.
- Both blocks were unconditional — no GPU guard, no version guard, nothing else.
- The two halves of `runtests.jl` are not a CPU/GPU split. The only top-level code between them is
  `@info` banners, so nothing depends on which block survives.
- History: `4b4084ad` added `"Test parallel inverses"` together with its `"Test parallel Cayley"`
  sibling. `7bf59f6f` added the duplicate, and took `test/arrays/triangular.jl` from 1 include to 2
  in the same commit. `8dc9b831` (2026-08-12) added the banners. All three claims verify.
- No reference to the removed testset name survives outside `CHANGELOG.md`.
- `fatou lint` clean, NFC clean, `test/runtests.jl` parses.

The choice to keep `"Test parallel inverses"` is sound. It sits next to `"Test parallel Cayley"`,
which is its real sibling.

## CI

Eleven of twelve checks pass. `Documentation` fails, and it is **pre-existing on `main`**, not
caused by this diff. That workflow has failed on every `main` push since 2026-08-26; the last green
run was `32851107456` on 2026-08-25. The `main` run right after #261 merged shows the same failure
class — missing `tikz/*.png` images and unresolved `@ref` anchors across `docs/src/**`, ending in
`makedocs encountered errors [:cross_references, :example_block]`. This branch changes no `docs/`
file, so it cannot have caused it.

## Nits — not fixed

- `CHANGELOG.md:270` is 101 characters against the bullet's 100-column wrap. The file holds lines up
  to 114, so this breaks no rule here.
- The PR description says `"Test parallel Cayley"` runs **60** tests. It runs **80** now, for the
  same reason as the 110.
- The surviving block sits under the `@info "Starting integrator and attention tests"` banner, while
  `kernels/tensor_mat_mul.jl` stays in the first half. The kernel tests stay split across both
  halves whichever block goes. Pre-existing, outside the diff, and not for this branch.

## Not checked

Type piracy, inference and allocations: no `src/` file changed and no callable symbol was touched.
Doctests: no `docs/` file changed. GPU backends: not available in this environment.

PR #264, opened 2026-09-14
Call default_parameters() in scripts/

michakraus:
`GeometricProblems`' `default_parameters` is a function, not a value. Fourteen call sites across
seven files under `scripts/` still used the bare name, so each one passed a `Function` where a
`NamedTuple` was required. This calls it.

A second commit then fixes the one call site where the `()` was necessary but not sufficient:
`H` in `scripts/ensemblesolution/harmonic_oscillator.jl` also had its arguments in the wrong order.
See *Follow-up* below.

Sub-task T14 of `Tasks/Restructure the GeometricMachineLearning test and script trees.md`.
Nothing outside `scripts/` and `CHANGELOG.md` is touched; `docs/src/tutorials/` was already correct.

## Pre-PR verification

This section records the pass that ran before the PR was opened, against the first commit only.
It is left as written. One defect it reported as pre-existing was fixed afterwards, under
*Follow-up*.

**Fixed in this pass:** nothing. No file was written.

**Unresolved:** none. No diff-attributable finding.

The 14 substitutions are correct. `default_parameters(::Type{T} = Float64) where {T}` is a
function in every problem module these scripts import — `HarmonicOscillator`, `DoublePendulum`,
`ABCFlow`, `RigidBody` — and in all three installed GeometricProblems versions (0.7.2, 0.8.3,
0.9.0). The 5th positional argument of `EnsembleProblem` is typed `::OptionalParameters`, which a
bare function cannot satisfy. The changelog's count of fourteen sites across seven files matches
the diff, and `docs/src/tutorials/` was indeed already correct.

**Pre-existing, adjacent — not fixed here.** Two of the seven scripts still cannot run. All three
defects are byte-identical on `main`.

| severity | file:line | claim |
|:---------|:----------|:------|
| bug | `scripts/symplectic_transformer/double_pendulum_short_integration.jl:115` | `const n_steps = 100` inside `make_plot_for_index`. `const` on a local is a lowering error, so the whole file fails to load and the fix at line 118 is unreachable. Reported by both `fatou lint` (`error: const-local`) and `Meta.lower`. Fix: drop `const`. |
| ~~bug~~ **fixed in `eb88d68`** | `scripts/ensemblesolution/harmonic_oscillator.jl:44` | The `hamiltonian` arguments are in the wrong order and the `p` slice is the whole vector, so the call is a `MethodError` with or without the `()`. Fix: `hamiltonian(0.0, x[1:(length(x) ÷ 2)], x[(1 + length(x) ÷ 2):end], default_parameters())`. |
| bug | `scripts/volume_preserving_feedforward/double_pendulum.jl:71` | `hodeproblem(…; params = params)` — the keyword is `parameters`. The value corrected at line 96 flows into it. |
| quality | `scripts/volume_preserving_feedforward/double_pendulum.jl:91` | `ω₀_val` is assigned and never read; line 92 uses `ω₀`. Same value, so no behaviour change. |
| correctness-risk | `test/training_phnn.jl:9,24,30` | The same defect class, in `test/`. The file is not included by `test/runtests.jl`, which is why CI is green. |

**Checked and clean**
- Correctness: all 14 hunks read against each callee signature in the resolved dependency source. No bare `default_parameters` remains as a value anywhere in `scripts/` or `docs/`; the remaining bare occurrences are import lines. `Float64` is right at every site.
- Callers: not applicable — the diff defines no function and changes no signature.
- Comments: the only comments in or adjacent to a hunk are present tense and describe current behaviour. None mentions `default_parameters`.
- History in prose: no `previously`/`used to`/`no longer`/version/date on any added line outside `CHANGELOG.md`.
- Formatting: the pre-commit hook's own JuliaFormatter command over the seven files returns ok.
- NFC and parse: both clean.

**Not checked**
- Type piracy: not applicable — the diff defines no method and touches no `src/` file. The repository has no Aqua configured, so there is no baseline either way.
- Type instability: no concrete call site. The changed lines are top-level script statements, and the scripts require Metal or CUDA hardware to load.
- Allocations: no number measured. No `@allocated` assertion or benchmark covers `scripts/`.
- Package load test and the test suite: no `src/`, `Project.toml` or `test/` file changed, so neither can be affected.
- Doctests: no `docs/` file changed.

## Verification

```
grep -rn 'default_parameters' scripts/ | grep -v 'default_parameters()' | grep -v using | grep -v import
```

returns nothing. Each touched script was then run in an isolated process with its own directory as
cwd: none fails on `default_parameters`. Both `rigid_body.jl` scripts now reach past the call site
and die on `CUDA driver not found`; the other five die *earlier*, on pre-existing unrelated causes
(`UndefVarError: timespan`, `Metal` not installed, `UndefVarError: timestep`). None of those was
fixed — they belong to later sub-tasks.

## Follow-up — the `H` argument order (`eb88d68`)

Raised in review, then fixed on this branch. `HarmonicOscillator.hamiltonian` is
`(t, q, p, params)`, and its array method forwards to `hamiltonian(t, q[1], p[1], params)`. `H`
passed the momentum slice as `t`, `0.0` as `q` and the whole state vector as `p`, so the call
matched no method. Calling `default_parameters()` was necessary but not sufficient: the line was
still a `MethodError`.

```julia
function H(x)
    hamiltonian(
        0.0, x[1:(length(x) ÷ 2)], x[(1 + length(x) ÷ 2):end], default_parameters())
end
```

**Measured**, in a throwaway environment holding `GeometricProblems` 0.8.3 — the version
`scripts/Manifest.toml` resolves — with the default parameters `(m = 1.0, k = 0.5)` at
`(q, p) = (0.3, 0.7)`:

| form | result |
|:---|:---|
| the old line | throws `MethodError` |
| the new line | `0.26749999999999996` |
| closed form `p² / 2m + k q² / 2` | `0.26749999999999996`, bit-identical |

`H([0.0, 0.0])` is `0.0`, and the grid call shape `plots.jl:86` uses —
`[H([x, y]) for x in X, y in Y]` — evaluates over a 5×5 sweep without error.

**Why a long-form definition.** JuliaFormatter under `style = "sciml"` rejects the single-line
assignment form: it is 95 characters against the 92 margin, and the block above is the formatter's
own rendering of the split. `format(f; overwrite = false)` returns `true` on the result. The
pre-commit hook passed all four stages — formatting, `fatou lint`, NFC, load.

**Still deferred**, unchanged from the table above: the `const n_steps` lowering error at
`double_pendulum_short_integration.jl:115`, the `params =` keyword at `double_pendulum.jl:71`, the
unread `ω₀_val` at `double_pendulum.jl:91`, and `test/training_phnn.jl`. They belong to later
sub-tasks.


michakraus (commented):
## Review — verified, ready to merge once CI is green

The change is correct, minimal and completely documented. Every claim in the PR body was
re-checked independently against the resolved dependency source. Nothing was found that this
branch introduces.

### Verified

**The substitution is right.** `default_parameters(::Type{T} = Float64) where {T}` is a function
in all four problem modules these scripts import, in the version the `scripts/` environment
resolves (`GeometricProblems` 0.8.3, `scripts/Manifest.toml`):

| module | definition |
|:--|:--|
| `HarmonicOscillator` | `src/harmonic_oscillator.jl:39` |
| `DoublePendulum` | `src/double_pendulum.jl:71` |
| `ABCFlow` | `src/abc_flow.jl:23` |
| `RigidBody` | `src/rigid_body.jl:27` |

The callees reject a `Function` at every one of the three call shapes:

- `EnsembleProblem`'s 5th positional argument is `parameters::OptionalParameters` or
  `AbstractVector{<:OptionalParameters}` — `GeometricEquations/src/problems/ensemble_problem.jl:73`.
  No constructor takes an untyped 5th argument, so the bare name was a `MethodError`.
- `odeproblem`/`hodeproblem`/`odeensemble` default their `parameters` keyword to
  `default_parameters()` themselves — e.g. `GeometricProblems/src/double_pendulum.jl:139`.
- `scripts/volume_preserving_feedforward/double_pendulum.jl:96` calls a local function whose 5th
  parameter is annotated `params::NamedTuple` (line 68). That one is decisive on its own.

`ϑ(t, q, q̇, params)` at `double_pendulum.jl:16,92` matches the upstream usage at
`GeometricProblems/src/double_pendulum.jl:81` exactly.

**The count and the completeness claim hold.** Running the sweep on both refs:

```
git grep -nE 'default_parameters([^(]|$)' <ref> -- scripts docs
```

`origin/main` returns the 14 sites across the 7 files named in the changelog; the branch head
returns only `using` lines. The positive control matters here — the filter does detect the defect,
so the empty result on the head is a real negative rather than a broken sweep.

**Formatting holds.** The seven head versions were checked with the pre-commit hook's own command
(`format(f; overwrite = false)`, `.JuliaFormatter.toml`, `style = "sciml"`) in an isolated
directory: 7 checked, none unformatted. The 112-character line at
`scripts/volume_preserving_transformer/rigid_body.jl:18` is over the margin but JuliaFormatter
accepts it, and over-margin lines are pervasive in `scripts/` on `main`.

**The changelog entry is in the right place** — `## [Unreleased] — 0.8.0`, `### Fixed` — and its
statements about the seven files and about `docs/src/tutorials/` are both accurate.

### Confirmed pre-existing, and not this branch's problem

Both defects the PR body reports were reproduced, and both are byte-identical on `main`:

- `scripts/symplectic_transformer/double_pendulum_short_integration.jl:115` — `const n_steps = 100`
  inside `make_plot_for_index`. `Meta.lower` returns
  `unsupported const declaration on local variable`, so the whole file fails to load and the fix at
  line 118 is unreachable.
- `scripts/volume_preserving_feedforward/double_pendulum.jl:71` — `hodeproblem(…; params = params)`.
  The keyword is `parameters` and the function takes no `kwargs...`, so the corrected value at
  line 96 flows into a call that still fails.

### One question, not a blocker

`scripts/ensemblesolution/harmonic_oscillator.jl:44` is the one line the diff edits that is still
guaranteed to fail after the edit. `hamiltonian` is `(t, q::AbstractArray, p::AbstractArray, params)`
(`GeometricProblems/src/harmonic_oscillator.jl:103`), so passing the `p` slice as `t`, `0.0` as `q`
and the whole of `x` as `p` matches no method. The `()` is necessary but not sufficient here.

Deferring it is defensible — it is a different defect, and the surgical rule says leave adjacent
code alone. But this is the same line and a one-token change away from correct. Worth folding in
here, or worth an explicit follow-up sub-task so it does not get lost.

`test/training_phnn.jl:9,24,30` carries the same defect class. The file is referenced nowhere in
`test/`, which is why CI never saw it; a later sub-task should either fix and include it or delete
it.

### CI

All ten checks were `pending` when this review was written — the run had just started. The diff
touches no `src/`, `test/`, `docs/` or `Project.toml` file, so no check can be affected either way,
but the matrix still has to come back green before merge.

### Verdict

Approve. No diff-attributable finding.

(`gh pr review --approve` is unavailable on an own-authored PR, so this is posted as a comment.)

michakraus:
## Fixed in `eb88d68` — the `H` argument order

Pushed the one-line fix I asked about above, plus its changelog entry.

```julia
function H(x)
    hamiltonian(
        0.0, x[1:(length(x) ÷ 2)], x[(1 + length(x) ÷ 2):end], default_parameters())
end
```

**Checked, not assumed.** In a throwaway environment holding `GeometricProblems` 0.8.3 — the
version `scripts/Manifest.toml` resolves — with the default parameters `(m = 1.0, k = 0.5)` at
`(q, p) = (0.3, 0.7)`:

| form | result |
|:--|:--|
| the old line | throws `MethodError` |
| the new line | `0.26749999999999996` |
| closed form `p²/(2m) + k q²/2` | `0.26749999999999996`, bit-identical |

`H([0.0, 0.0])` is `0.0`, and the grid call shape `plots.jl:86` uses — `[H([x, y]) for x in X, y in Y]`
— evaluates over a 5×5 sweep without error.

**Why it is a long-form definition.** JuliaFormatter under `style = "sciml"` rejects the
single-line assignment form: it is 95 characters against the 92 margin, and the formatter's own
rendering of the split is the `function … end` block above. `format(f; overwrite = false)` returns
`true` on the result, and the pre-commit hook passed all four stages (formatting, `fatou lint`,
NFC, load).

**Two rows of the PR body are now stale.** "Fixed in this pass: nothing. No file was written." no
longer holds, and the `scripts/ensemblesolution/harmonic_oscillator.jl:44` row of the
pre-existing-defects table is now fixed rather than deferred. The other two rows still stand: the
`const n_steps` lowering error at `double_pendulum_short_integration.jl:115` and the
`params =` keyword at `double_pendulum.jl:71` are untouched, as is `test/training_phnn.jl`.

I authored this fix, so the merge is not mine to make.

PR #269, opened 2026-09-15
Guard every test file against being unreachable

michakraus:
`test/` holds 102 `.jl` files. **62 are reachable from `runtests.jl`; 40 are not.** A test file that
nothing includes is never run and nothing says so: it keeps compiling in a reader's head, it gets
repaired by hand from time to time, and it regresses in silence. `git log` shows exactly that
happening.

`test/reachability.jl` walks the transitive `include` closure of `runtests.jl` and requires every
file under `test/` to be either inside that closure or named in an allowlist with a reason. Adding a
test file without wiring it in now fails the suite, and a file deliberately not run has to say why.

**Every reason states what running that file actually does**, measured — not guessed from its name.
All 40 were run individually in the test environment. The groups are the observed first blocker:

| group | count |
|:--|--:|
| stop at `using` a package that is neither a dependency nor a test target | 23 |
| name something the package does not provide | 13 |
| run and assert something | 2 |
| load without error and assert nothing | 2 |

That last group is the one worth knowing about: `gram_schmidt.jl` defines two test functions and
calls neither, so "it runs" says nothing about whether it works.

The closure is read off the parsed syntax tree, not the text. A line scanner has to re-implement
Julia's lexer to tell whether an `include` it found is code, and gets `#= … =#` and string literals
wrong — an earlier line-based version of this guard passed vacuously when an `include` was placed
inside a block comment.

This closes open issue **C11**, whose count of 41 was wrong: `macro_testerror.jl` and the `data/`
files are reachable transitively through `data/test_batch.jl` and `data/test_matching.jl`.

Sub-task T1 of `Tasks/Restructure the GeometricMachineLearning test and script trees.md`.

## Pre-PR verification

**Fixed:** a Windows path-separator bug in `test/reachability.jl` (`relpath` emits `\`-separated paths on `windows-latest`, a required CI matrix job, which would have failed the new orphan-reachability testset on every Windows run by making all 40 legitimate `ALLOWED_ORPHANS` entries look unallowlisted) — normalized to `/` at both relative-path sites; also reworded one `ALLOWED_ORPHANS` reason (`training_phnn.jl`) off historical narration ("which GeometricProblems made a function") onto present-tense fact, per house style on comments outside `CHANGELOG.md`.

**Unresolved:** none.

**Pre-existing, adjacent:** none.

**Checked and clean:** the AST-walk closure logic (multi-level transitive includes, commented-out includes correctly ignored, files outside `test/` correctly dropped); the CHANGELOG's `102 = 62 reachable + 40 unreachable` claim and its full category breakdown, independently recomputed and matched exactly; the `### Infrastructure` heading is one the file already uses, under the existing `## [Unreleased]` section; no other present-tense/historical-language issues in the diff.

**Not checked:** type piracy (no methods on foreign types — N/A), type instability and allocations (test-only scaffolding, no performance claim — N/A).

## Verification

`Pkg.test()` passes. The guard was demonstrated to detect a positive three ways, each reverted:
removing one allowlist entry while leaving its file in place fails naming that file; adding an
unreferenced `test/scratch_probe.jl` fails naming it; and an `include` placed inside a `#= … =#`
block comment is correctly **not** treated as live. A reviewer independently re-derived the closure
with its own implementation (102 files, 62 reachable, 40 orphans, exact match) and ran 13
allowlisted files individually to check their stated reasons.


michakraus:
## Review — posted after the merge, for the record

The guard is right, and the part easiest to get wrong is the part it gets right: the closure is read
off `Meta.parseall`, so a commented-out or string-literal `include` is not live without the guard
re-implementing Julia's lexer. The `\` → `/` normalisation at both `relpath` sites is what keeps the
three `windows-latest` jobs honest.

I re-derived the closure independently against the merged tree: **80 `.jl` files under `test/`, 63
reachable, 17 orphans, 17 allowlist entries, nothing unlisted.** Exact match. Against the tree before
#270 landed it was 103 / 63 / 40, which is the figure the *Infrastructure* entry states, and 103 − 23
= 80 and 40 − 23 = 17 line up with the entry #270 added beneath it.

### One finding, fixed on the branch before it merged

**`ALLOWED_ORPHANS` was closed in one direction only.** It stopped a test file from going unlisted.
It did not stop an *entry* from outliving the file it excuses: wire a file into `runtests.jl`, or
delete it, and the entry stayed behind carrying a reason that no longer held, with nothing to say so.

That is the direction this allowlist gets edited in. It is a backlog of files waiting to be repaired,
wired in or deleted, so a stale entry is the likely failure, not an unlisted file — #270 pruned 23
entries alongside its deletions and #271 prunes 9 more, both by hand. `test/exports.jl`, the sibling
guard from #266, already closes its allowlist both ways with a second `setdiff`.
`test/reachability.jl` now does the same:

```julia
stale = filter(f -> f in REACHABLE || !(f in files), sort!(collect(keys(ALLOWED_ORPHANS))))
@test stale == String[]
```

Demonstrated to detect a positive in both flavours: an entry naming a file that is now included, and
an entry naming a file that no longer exists. Both are reported by name. *C11* and the
*Infrastructure* entry record it.

### Merge

`main` had gained `test/exports.jl` from #266, which claimed the same first slot in `runtests.jl`.
Both testsets are kept, reachability first. The `CHANGELOG.md` halves auto-merged.

### CI

`0b3c94d1`, the commit that merged, is **green on all nine jobs** — the seven required contexts
(`Julia 1` and `Julia min` across ubuntu / macOS / windows, plus `Doctests - ubuntu-latest`) and the
non-required `nightly` and `pre`. `main` at `3bccaea9` carries a byte-identical tree.

`Documentation` is **not** a required check and is **pre-existingly red** on every recent branch:
a `MethodError: no method matching Optimizer(::GradientMethod, ::@NamedTuple{xy::Matrix{Float64}};
step_size)` in the `docs/src/tutorials/optimizer_comparison.md` `@example` block, plus a long list of
`invalid local link/image` errors. Nothing in this branch touches `docs/`; it needs its own change.

### Not checked

Type piracy (no methods on foreign types — N/A). Type instability and allocations (test-only
scaffolding, no performance claim — N/A).

PR #273, opened 2026-09-15
Convert two verification scripts into tests

michakraus:
`scripts/test_attention.jl` and `scripts/test_double_multiplication_derivative.jl` both failed on
`FieldError: type NamedTuple has no field params`. They are now testsets under `test/`.

**The cause is an API trap worth knowing about.** `NetworkParameters` overloads `getproperty`, so
`ps.params` indexes `:params` as a `NamedTuple` key instead of returning the wrapper. The accessor
is `params(ps)`.

**What these scripts actually checked was not a derivative identity.** They probe *structure
preservation* through Zygote: whether a gradient taken with respect to a `NetworkParameters`
wrapper comes back with its `SymmetricMatrix` structure intact, versus one taken with respect to
the bare `NamedTuple`. Their own comments marked some paths "this doesn't work" and others "this
does work" — that contrast is the content, so it is asserted rather than tidied away:

- Where a parameter is used **once**, the structure survives the wrapper. Asserted.
- Where it is used **twice**, it comes back a plain `Matrix`. Recorded as `@test_broken`, not deleted.

The tests also assert something neither script stated, which is the real mathematical content: the
gradient through the wrapper **equals** the one through the bare `NamedTuple`. Measured difference
is exactly `0.0` in every case, including the cases that lose structure — so the defect is
structural, not numerical.

The cause of the structure loss is stated only as far as it was established. An earlier draft blamed
`_custom_mul` lacking an `rrule`; a control experiment disproved it — `gradient(p -> sum(p.L1.A) +
sum(p.L1.A), ps)`, with no `_custom_mul` anywhere, loses structure through the wrapper and keeps it
through `params(ps)`. The entry now attributes it to the wrapper's own gradient accumulation across
repeated uses of a leaf, and says plainly that the internal mechanism was not isolated further.

Sub-task T13 of `Tasks/Restructure the GeometricMachineLearning test and script trees.md`.

## Pre-PR verification

**Fixed:** `test/runtests.jl:86` — new `@safetestset` name string was 79 chars, one short of the 80-char column convention the rest of the file follows; padded to 80.

**Unresolved:** none blocking. Nit (not fixed, needs a design decision): `network_parameters_gradient_structure_test` is duplicated verbatim across the two new test files rather than shared via a common helper.

**Pre-existing, adjacent:** none found.

**Checked and clean:** correctness/no dangling references to the deleted scripts; no type piracy (new names only, no Aqua baseline in this package); `Zygote.gradient`'s `Any`-typed return is confirmed pre-existing/inherent (control case outside the diff shows the same instability), not a regression; comments/docstrings present-tense and accurate; no history language outside `CHANGELOG.md`; CHANGELOG entry present under `[Unreleased]`; import hygiene clean, all deps already declared; `@safetestset` wiring matches convention; JuliaFormatter-clean; NFC-clean; `fatou lint` clean.

**Not checked:** avoidable allocations (no `@allocated` assertions in the diff); Aqua tool run (package has no Aqua dependency); full `Pkg.test()` (established clean by prior review rounds, not re-run per instructions).

## Verification

`Pkg.test()` passes. Both testsets show exactly the expected one `Broken` each — not silently fixed,
not silently erroring: `Symplectic attention: NetworkParameters gradient structure | 11 1 12` and
`_custom_mul: NetworkParameters gradient structure | 5 1 6`.

A reviewer confirmed independently that the two gradients are genuinely separate Zygote calls
(`g_wrapped !== g_bare`) rather than a self-comparison, that each `@test_broken` would fail if
written as `@test`, that no case from the original scripts was dropped, and reproduced the control
experiment that fixes the cause.


michakraus (commented):
## Review — Convert two verification scripts into tests

**Verdict: approve after one documentation correction.** The conversion itself is faithful and
correct. I reproduced every claim the body makes. What needs changing is the *explanation* the
`CHANGELOG.md` entry and the two docstrings give for the structure loss: the stated rule is
falsified by this PR's own layer testset, and the mechanism the entry says "was not isolated
further" turns out to be isolable in one line.

Reviewed from a detached worktree at `origin/pr/273` (`d15d7334`).

---

### Findings

| 1 | correctness-risk | `CHANGELOG.md:446-450`; `test/attention_layer/symplectic_attention_network_parameters_gradient.jl:12-14` | The stated rule — structure survives when the leaf is "used once" and degrades when "used twice" — is not the rule. The predictor is the number of **`getproperty` accesses on the wrapper**, not the number of uses of the value. | control below: binding `A = _ps.L1.A` once and using it **twice** preserves `SymmetricMatrix`; two `_ps.L1.A` accesses lose it. |
| 2 | quality | `test/attention_layer/symplectic_attention_network_parameters_gradient.jl:74-77` | The `SymplecticAttentionQ` testset asserts `preserves_structure = true`, but the layer's forward pass uses `A` **twice** (`src/layers/symplectic_attention.jl:129,131`). Under the rule as written it should be `@test_broken`. It passes only because the layer binds `A` to a local first. Nothing in the file or the CHANGELOG says so, so this case reads as an unexplained exception. | `src/layers/symplectic_attention.jl:128`; run output below. |
| 3 | quality | deleted `scripts/test_attention.jl:32` vs new file `:77` | The PR body says the scripts' "this works" / "this doesn't work" contrast "is the content, so it is asserted rather than tidied away". For the layer case the contrast was **inverted**, not preserved: the script commented `∇₃ = gradient(params -> norm(nn(t, params)), ps)[1] # this doesn't work`, and the new test asserts it *does* work. The re-classification is correct — I reproduced it — but it is silent. | measured: `orig-script-style ps A type: SymmetricMatrix{Float64, Vector{Float64}}`. |
| 4 | quality | `test/attention_layer/…:21-37` and `test/custom_ad_rules/…:17-33` | `network_parameters_gradient_structure_test` is duplicated verbatim, 17 lines, across two files. The body flags this as an open nit. `test/reachability.jl` does not block a shared helper: both includers are in `runtests.jl`'s transitive closure, so a `test/network_parameters_gradient_structure.jl` included by both is reachable. | `test/reachability.jl:44-56` walks the parsed `include` closure. |
| 5 | nit | both new files, line 10 | `@doc raw"""` — neither docstring contains a backslash or a `$`, so `raw` buys nothing. | read both docstrings. |
| 6 | nit | `test/attention_layer/…:77` | `nn.params`. The lesson this very file exists to record is that `.params` is a trap; the accessor is `params(…)`. It is safe here — `nn` is a `NeuralNetwork` with a real field, not a `NetworkParameters` — but it reads against the file's own point. | `params(nn) === nn.params: true`. |

---

### The control that isolates finding 1

Fresh Julia 1.13 process, package environment via `TestEnv`, `--startup-file=no`:

```julia
S = rand(SymmetricMatrix, 4); ps = NetworkParameters((L1 = (A = S,),)); a = rand(4, 4)

dm_two(a, _ps) = _custom_mul(_ps.L1.A, _custom_mul(a, _ps.L1.A))   # two getproperty accesses
function dm_one(a, _ps)                                             # one access, value used twice
    A = _ps.L1.A
    _custom_mul(A, _custom_mul(a, A))
end
```

```
two accesses                           -> Matrix{Float64}
one access, local bound, used twice    -> SymmetricMatrix{Float64, Vector{Float64}}
sum: two accesses                      -> Matrix{Float64}
sum: one access used twice             -> SymmetricMatrix{Float64, Vector{Float64}}
```

The last two rows are the PR body's own `sum(p.L1.A) + sum(p.L1.A)` control, split the same way.
So the loss is not "the wrapper's gradient accumulation across the repeated use" of the value —
the value is used twice in both rows. It is the accumulation of **two `getproperty` pullbacks on
the wrapper into one leaf**.

That also answers a pre-existing `TODO` the PR sits right next to:

```julia
# src/layers/symplectic_attention.jl:128
A = ps.A # for some reason we have to allocate a local variable here. This should be further investigated.
```

That local binding is exactly the workaround, and this PR now has the experiment that explains
it. Since this is the documented cause of the two `@test_broken`s, it is worth folding the answer
into the same change: it turns two open mysteries into one stated rule, and it makes the layer
testset (finding 2) consistent rather than exceptional.

**Suggested:** reword the `CHANGELOG.md` mechanism and both docstrings from "used once / more
than once" to "reached through one `getproperty` on the wrapper / more than one"; add the
`dm_one` case above as a third testset in
`test/custom_ad_rules/double_multiplication_network_parameters_gradient.jl`, so the rule is
asserted rather than only asserted-about; and update `src/layers/symplectic_attention.jl:128`'s
comment now that "for some reason" has a reason.

---

### Verified and fine

- **Both testsets run and produce exactly the claimed result.** Fresh process, package
  environment: `Test Summary: PR273 | 16 pass, 2 broken, 18 total, 15.4s`. That matches the
  body's `11 1 12` plus `5 1 6`.
- **Each `@test_broken` genuinely fails as a plain `@test`.** Measured:
  `double_multiplication wrapped A type: Matrix{Float64}` and
  `symplectic_attention wrapped A type: Matrix{Float64}`. Neither is a silently-passing
  `@test_broken`.
- **The two gradients are separate Zygote calls, not a self-comparison.** `g_wrapped !== g_bare`
  is `true`.
- **The `isapprox` claim is exact.** `maximum(abs.(Matrix(a_wrapped) .- Matrix(a_bare)))` is
  `0.0` for the layer case. The structure loss is cosmetic, as the body says.
- **No case from the original scripts was dropped.** All five functions carry over:
  `symplectic_attention`, `symplectic_attention_simplified`, `symplectic_linear_map`,
  `single_multiplication`, `double_multiplication`, plus the `SymplecticAttentionQ` layer.
- **The `ps.params` diagnosis is right.** `NetworkParameters` overloads `getproperty` into the
  wrapped `NamedTuple`, so `ps.params` raises `FieldError`. `params` is imported at
  `src/GeometricMachineLearning.jl:92` and not exported, which is why the test files reach it
  with `using GeometricMachineLearning: params`.
- **Dimensions are right, and not by luck.** `SymplecticAttentionQ(4)` gives a 2×2 `A` (`M = 4`
  is the full phase-space dimension), so `t = (q = rand(2,2), p = rand(2,2))` is correct. A
  4-row `t` fails with `AssertionError: A.n == size(B, 1)`.
- **No dangling references to the two deleted scripts** anywhere outside `CHANGELOG.md` (grep
  over `*.jl`, `*.md`, `*.yml`, `*.toml`).
- **`test/reachability.jl` needs no change**, as claimed — both new files are in `runtests.jl`'s
  include closure, so neither is an orphan.
- **The `runtests.jl` column convention holds.** Both new `@safetestset` names are exactly 80
  characters, matching the 58-of-63 majority.
- **No type piracy.** The diff adds six function names, all new, all inside `@safetestset`
  modules. No method is added to a foreign type.
- **Imports clean.** `_custom_transpose` is used (line 41); `_custom_mul`, `params`, `norm`,
  `gradient` and `Random` are all used. No stale import in either file.
- **JuliaFormatter-clean** under the repo's `style = "sciml"`: `format_file(…; overwrite=false)`
  returns `true` for both new files and for `runtests.jl`.
- **NFC-clean**: `s == Unicode.normalize(s, :NFC)` holds for both new files.
- **CHANGELOG entry present** under `## [Unreleased] — 0.8.0`, section `### Added`, line 439.
  Present tense, and no history language leaks into the source files.
- **The task's own acceptance criterion is met.** T13 asks for `@test_broken` on the non-working
  cases *plus* an assertion that the wrapped and unwrapped paths agree numerically. Both are
  there.

### Not checked

- **Allocations and type stability** — the diff is test-only and adds no `@allocated` assertion.
  `Zygote.gradient`'s `Any` return is inherent, and the body's claim that it is pre-existing
  holds by inspection: the control expressions outside the diff have it too.
- **Aqua** — the package declares no Aqua dependency, so there is no piracy or ambiguity
  baseline to run against. The "no type piracy" line above is by inspection of the diff, not by
  tool.
- **Full `Pkg.test()`** — I ran the two new files in isolation against the package environment
  rather than the whole suite. CI covers the rest.

### CI

All 7 required checks pass, read job by job rather than from the workflow conclusion:

```
SUCCESS   Doctests - ubuntu-latest              (required)
SUCCESS   Julia 1   - ubuntu / macOS / windows  (required)
SUCCESS   Julia min - ubuntu / macOS / windows  (required)
SUCCESS   Julia nightly - ubuntu-latest         (not required, and green anyway)
SUCCESS   Julia pre     - ubuntu-latest         (not required)
SUCCESS   codecov/patch, codecov/project
IN_PROGRESS  Documentation                      (not required)
```

Nothing is red. Nightly is green here, so it is not the usual expected-red case. Only
`Documentation` is still running, and it is not a required check.

PR #274, opened 2026-09-15
Correct the gradient structure rule from #273 and share its assertion helper

michakraus:
Follow-up to #273, which merged before its review findings were addressed. Test and CHANGELOG only; no `src/` line changes.

## What was wrong

#273 recorded this rule: a `SymmetricMatrix` leaf survives a Zygote gradient when the leaf is **used** once in the loss expression, and degrades to a plain `Matrix` when it is used twice.

That is not what decides it, and there are two routes in that count at different levels.

**Through `getproperty` on the wrapper**, the count is of accesses on the `NetworkParameters` wrapper:

```julia
gradient(p -> (A = p.L1.A; sum(A) + sum(A)), ps)  # -> SymmetricMatrix   one access, two uses
gradient(p -> sum(p.L1.A) + sum(p.L1.A), ps)      # -> Matrix            two accesses, two uses
```

Two uses through one access keep the structure, a third access behaves as the second, and accesses on the plain `NamedTuple`s below the wrapper are free. The two readings coincide in every case the tests cover, which is what let the weaker one stand.

**Through `values(ps)`**, which is how `Chain` reaches a layer (`AbstractNeuralNetworks/src/chain.jl:56`), the accesses on that inner `NamedTuple` are what decide, and the wrapper count no longer shields them:

```julia
gradient(p -> (L = values(p)[1]; sum(L.A) + sum(L.A)), ps)  # -> Matrix
```

The bare wrapped `NamedTuple` (`params(ps)`) keeps the structure on either route at any access count.

## What that settles

The `SymplecticAttentionQ` testset asserts the structure survives although the layer body reads `A` twice. It takes the `values` route, so the local binding at `src/layers/symplectic_attention.jl:128` — whose comment calls itself unexplained — **is** load-bearing: an A/B toggling only the binding, same closure, same input, same route, returns a `SymmetricMatrix` with it and a plain `Matrix` without it.

The comment in `src/` is left in place. This measures structure preservation only, and says nothing about whatever else the binding may have been added for.

## Also in this PR

- **The inverted script comment is now recorded.** #273 flipped the `SymplecticAttentionQ` case from the deleted script's `# this doesn't work` to an asserted pass, silently. It does work — both the script's own form and the testset's return a `SymmetricMatrix` — but the inversion belonged in the CHANGELOG.
- **The duplicated assertion helper is shared.** 17 lines were copied verbatim into both test files; they now include `test/network_parameters_gradient_structure.jl`. A `@safetestset` is its own module, so sharing means including. The rule the helper documents is the one thing that must not drift between two copies.
- Plain docstrings instead of `@doc raw` with nothing to escape; testset names that say accesses rather than uses; `params(nn)` instead of `nn.params`, against which these very files are the cautionary tale.

## Pre-PR verification

**Fixed during verification** — the first draft of this branch claimed `Chain` performs `_ps.L1` and that the local binding is *not* load-bearing. `julia-branch-verifier` falsified both against `chain.jl:56` and the binding A/B. The comment, the helper docstring and the CHANGELOG now state the measured mechanism.

**Checked and clean**
- The access-count rule reproduced in a fresh process on both routes.
- `params(nn) === nn.params`, both `NetworkParameters`.
- The shared helper's include resolves from both test directories.
- The reachability guard passes with no allowlist change.
- `fatou lint`, `nfc.jl` and JuliaFormatter clean; the pre-commit hook passes all four stages.

**Not checked** — type piracy, type instability and allocations: the diff changes no `src/` line. Doctests: `docs/` untouched.

**Tests** — 16 passed, 2 broken, 18 total, unchanged from #273. The two `@test_broken` cases are the two-access ones.

## Note for the reviewer

The remote branch `T13-followup-review-fixes` holds a superseded first draft of this change and has no PR. It can be deleted.


michakraus (commented):
## Review

The code, the tests and the helper sharing are sound, and every empirical claim in the PR body reproduces in a fresh Julia 1.13 process against the head. One prose claim did not, and I have pushed the correction as `c6e0660e` rather than leaving it for a second round.

### The one finding

The observational rule is right: the split is decided by the number of `getproperty` calls on the wrapper, not by how often the leaf is used, and the `values` route counts at a different level. The **causal** half was wrong. `CHANGELOG.md` attributed the loss to the wrapper's own gradient accumulation and said the bare `NamedTuple` keeps the structure at any access count; the helper's docstring repeated the second clause.

The reverse pass loses the structure identically for both. What differs is the projection that runs afterwards:

```
pullback, one access,   NamedTuple        => SymmetricMatrix{Float64, Vector{Float64}}
pullback, one access,   NetworkParameters => SymmetricMatrix{Float64, Vector{Float64}}
pullback, two accesses, NamedTuple        => Matrix{Float64}
pullback, two accesses, NetworkParameters => Matrix{Float64}
gradient, two accesses, NamedTuple        => SymmetricMatrix{Float64, Vector{Float64}}
gradient, two accesses, NetworkParameters => Matrix{Float64}
ProjectTo(NamedTuple) applied to a Matrix => SymmetricMatrix{Float64, Vector{Float64}}
ProjectTo(NetworkParameters)              => typeof(identity)
```

`Zygote.gradient` finishes by projecting its result, and `ChainRulesCore.ProjectTo` of a `NamedTuple` is a structured projector with a `ProjectTo{SymmetricMatrix}` leaf that maps the raw `Matrix` back. `ProjectTo` of a `NetworkParameters` falls back to `identity`, because `NeuralNetworkParameters` defines no method for it, so nothing restores the wrapper's leaf. Defining that method upstream would clear both `@test_broken`s.

`c6e0660e` corrects the changelog paragraph and the docstring clause, and archives the check as `scripts/network_parameters_gradient_projection.jl`, so the claim now in the changelog has a runnable check next to it.

### Verified and fine

- The access-count rule, the `values`-route split and the bare-`NamedTuple` results all reproduce exactly as the PR body states.
- The `values` citation is exact: `AbstractNeuralNetworks/src/chain.jl:56` is `@inline applychain(layers::Tuple, x, ps::NetworkParameters) = applychain(layers, x, values(ps))`.
- The "load-bearing local binding" claim holds under an A/B on the symmetric body of `src/layers/symplectic_attention.jl:128`: with the binding the leaf is a `SymmetricMatrix`, without it an `Array`, and the two agree numerically.
- Tests: `_custom_mul` 5 pass / 1 broken, `Symplectic attention` 8 pass / 1 broken, `SymplecticAttentionQ layer` 3 pass. 16 pass, 2 broken, 18 total — unchanged by the correction.
- The reachability guard passes with no allowlist change; the shared helper is reached through the transitive `include` closure.
- No orphaned imports after the move. SciML formatting clean, NFC clean, `fatou lint` clean.
- No `src/` lines change, so type piracy, inference and allocation checks do not apply to this diff.
- Sharing the helper by `include` rather than duplication is the right call, and the relative path resolves from both test directories.

### CI

All 10 `CI` matrix jobs passed on the previous head, including `nightly` and the three `min` jobs, as did both codecov checks. `Documentation` is red, but the failure is pre-existing and outside this diff — an `@example` block in `docs/src/tutorials/optimizer_comparison.md:3` plus invalid local links. The same job is red on #273 and #272, and this PR changes no file under `docs/`.

### Verdict

Approve, subject to the re-run on `c6e0660e` coming back green.

PR #275, opened 2026-09-15
Wire in two surviving test orphans

michakraus:
Wires in the two test orphans worth keeping, and the repair turned up something.

`test/layers/sympnet_upscaling.jl` named `GradientQ`/`GradientP`, which do not exist — the layers
are `GradientLayerQ`/`GradientLayerP` — and its loop called `test_symplecticity()` with no
arguments, so `N` and `N2` were swept and then ignored while one default case ran 100 times. It now
runs **120 distinct `(N, N2)` pairs**. `test/transformer_related/transformer_setup.jl` already
passed and is wired in unchanged. Both allowlist entries are removed, 8 down to 6.

## The upscaling composition is not exactly symplectic

Repairing a test that had never run exposed a false premise underneath the one it was written for.

The original assertion was that the whole chain — PSD-encode `N→N2`, shear, PSD-decode `N2→N` — is
symplectic. Unseeded, it fails **12.7% of draws at `N=2`** (150 trials), 2.7% at `N=4`, 0% at `N≥6`.

**This is algebraic, not numerical.** Three things rule out the usual explanations:

- Identical trials in **Float64** give the same failure rate (11.7% both) and the same error
  magnitudes (mean relative error 0.052 vs 0.057). A rounding artefact would have shrunk by nine
  orders of magnitude.
- `cond(Φ)` is **exactly 1.0000** in every trial, pass or fail — Φ comes from `qr!(...).Q` and is
  orthonormal by construction.
- The mechanism: the round trip requires the shear layers to preserve `P = E𝕁ₙE'`, which is
  rank-deficient (rank `N` < `N2`), while they only guarantee the full-rank `𝕁_{N2}`. Those cannot
  be equal. Round-trip error runs 0.011–0.275, shrinking with `N` but with no reason to vanish at
  finite `N2`.

Each individual layer *is* exactly symplectic, to ~1e-16. So the test now asserts the three
identities that hold exactly and does not compute the round trip at all:

| identity | max deviation measured |
|:--|--:|
| encoder `E'𝕁_{N2}E = 𝕁ₙ` | 7.6e-7 |
| shear `G'𝕁_{N2}G = 𝕁_{N2}` | 4.3e-7 |
| decoder `Dec·𝕁_{N2}·Dec' = 𝕁ₙ` | ~5e-7 |

all against `atol = 1e-5`. A first attempt pinned the flakiness with `Random.seed!(1234)`; that was
rejected, because the seed did not stabilise an ill-conditioned draw — it concealed a false claim.
There is no seed in the file.

**Whether this architecture is *meant* to be exactly symplectic here is an open question**, and this
PR does not answer it. It changes only what the test claims.

Two smaller findings came with it. `PoissonTensor(N÷2)` was a second latent bug — the argument is
the full dimension, confirmed by tracing `PSDLayer(N,N2) → GradientLayerQ/P(N2) → PSDLayer(N2,N)`,
which maps `R^N → R^N`. And `test/runtests.jl:41` labelled `layers/sympnet_layers_test.jl` "Test
symplecticity of upscaling layer" though that file contains no symplecticity check; it is now
labelled for what it does, and `sympnet_upscaling.jl` carries the symplecticity claim.

Sub-task T12 of `Tasks/Restructure the GeometricMachineLearning test and script trees.md`.

## Cost

The testset runs 480 assertions in ~2m50s. An ablation attributes ~48s of that to the decoder
Jacobian alone; the cost is the Zygote Jacobians through the nonlinear shear layers, not the
comparisons. Dropping the round-trip gate did **not** make it cheaper.

## Verification

`Pkg.test()` passes. 200 unseeded trials at the smallest swept case `(N,N2)=(2,4)` gave zero
failures on every assertion, so the flakiness is genuinely gone rather than pinned.

A reviewer independently derived that `Dec·𝕁_{N2}·Dec'` is the only dimensionally valid form for an
`N2→N` map and confirmed it is the correct pushforward-of-bivector identity — not merely a form
that happens to pass — and re-measured its deviation over 1,300 unseeded draws.

Checked and clean: JuliaFormatter (SciML style), NFC, `fatou lint`, import hygiene, and
`@safetestset` name padding matching the file's 80-column convention.


michakraus (commented):
## Review — PR #275, *Wire in two surviving test orphans*

**Verdict: approve in substance.** The three claims that matter — `PoissonTensor` takes the full
dimension, the round-trip composition is not exactly symplectic, and the three per-layer identities
are — all reproduce here independently. Two things are worth a commit on this branch before merge:
the sweep is expensive for what it buys, and the dead default arguments reinstate the exact footgun
this PR repairs.

All measurements below were made on a detached worktree at `73ee4e8a`, fresh process each time,
`--project=.`, unseeded RNG.

### CI

All nine test jobs **pass** (run `34994702136`): Julia `1`, `min` and `pre` across ubuntu, macOS and
windows, plus *Doctests*. `codecov/patch` and `codecov/project` pass. *Julia nightly* and
*Documentation* were still running when this was written — nightly is expected to be red here and
is not a required check.

### Findings

| 1 | quality | `test/layers/sympnet_upscaling.jl:37-41` | 120 `(N, N2)` pairs cost **4m43.3s per CI job** under `Pkg.test()`'s `--check-bounds=yes`, and the sweep buys no information — the measured deviation is flat across the whole range | `--check-bounds=yes`: `480 passed, 4m43.3s`. `--check-bounds=auto`: `2m41.2s`. The local figure is the reliable one — same machine, same process shape. CI durations are consistent with it but too noisy to attribute cleanly (two `main` runs an hour apart, `34974596166` and `34971394598`, disagree by 10 min on *Doctests* alone), so I do not quote a matrix-wide total. Max deviation over 40 unseeded draws at each of five corners: `(2,4)` 3.6e-7 · `(2,8)` 3.6e-7 · `(10,40)` 4.8e-7 · `(20,40)` 5.4e-7 · `(20,80)` 8.3e-7. A dozen pairs would show the same thing |
| 2 | quality | `test/layers/sympnet_upscaling.jl:8` | the defaults `N = 4, N2 = 20` are now dead and reinstate the footgun the PR fixes | every call site (line 39) passes both arguments. The original bug was exactly that `test_symplecticity()` with no arguments silently ran one default case 100 times. Dropping the defaults turns a repeat of that mistake into a `MethodError` |
| 3 | quality | `test/reachability.jl:15-19` | the PR removed the only allowlist entry describing a file that *works*, and orphaned the header comment that describes it | "The **groups** are the observed first blocker" — there is now one group (line 22). "a file that is said to **work** was seen to exit cleanly" now describes no entry; all six remaining reasons are errors |
| 4 | nit | `test/layers/sympnet_upscaling.jl:34` | the fourth assertion is a matrix-column-consistency check, not a symplecticity check, and it duplicates the sibling file this PR just relabelled | the testset is newly named "Test symplecticity of the sympnet upscaling layer"; line 34 asserts `model(ten, ps)[:, 1] ≈ model(ten[:, 1], ps)`. `test/layers/sympnet_layers_test.jl:32-34` asserts the same shape of claim and is relabelled "tensor-slice consistency" by this very PR. 120 of the 480 assertions are therefore off-topic for the name |
| 5 | nit | `CHANGELOG.md` (new entry) | "measured max deviation 2–5e-7 across `N`" understates, and disagrees with the PR body | the PR body's own table says 7.6e-7 (encoder). I measure **8.34e-7** for the shear identity at `(N,N2) = (20,80)` over 40 draws. Headroom against `atol = 1e-5` is ~12x, not ~20x |
| 6 | nit | `test/layers/sympnet_upscaling.jl:7` | "see `CHANGELOG.md` for the measured round-trip error" points at a 900-line file with no anchor | name the `[Unreleased] — 0.8.0` section, or inline the range (`0.011`–`0.275`), which is one clause |
| 7 | nit | `test/layers/sympnet_upscaling.jl` (deleted line) | the deleted comment "the first and last PSD layer need to have the same weight! (else they map to a different symplectic potential)" stated a true and non-obvious fact that now appears nowhere — not in the test, not in the CHANGELOG | measured: median round-trip `‖J'𝕁J − 𝕁‖∞` over 40 draws at `(2,4)` is **4.3e-2 tied** vs **5.8e-1 untied**; at `(20,80)` **7.0e-3 tied** vs **8.7e-1 untied**. Tying is worth an order of magnitude. Dropping the line is right — nothing asserts the round trip any more — but the fact is worth one line in the CHANGELOG |

### Verified and fine

- **`PoissonTensor(N÷2)` really was a bug.** `src/arrays/poisson_tensor.jl:43-51` sets `n = n2÷2`
  and builds a `2n×2n` matrix, so the argument is the full dimension. The docstring at line 3-5
  agrees. `PoissonTensor(N)` is correct for an `N×N` Jacobian.
- **The three identities hold, with headroom, unseeded.** 200 draws over five `(N, N2)` corners
  spanning the sweep: **0 failures**, max deviations `E` 5.96e-7, `G` 8.34e-7, `Dec` 5.96e-7
  against `atol = 1e-5`. The flakiness is genuinely gone rather than pinned — there is no seed in
  the file.
- **The round-trip claim survives the removal of the weight tying.** This was the check worth
  making: the PR drops `ps[4].weight.A = ps[1].weight.A`, so "the round trip is only approximately
  symplectic" could have been measuring an untied chain. It is not. With the tying **restored**,
  `‖J'𝕁J − 𝕁‖∞` runs 7.0e-3 to 5.1e-1 over 40 draws at each of five corners, and is above `1e-5`
  in **200 of 200** draws. Not asserting the round trip is the right call.
- **The runtime figure reproduces.** Claimed `2m44.2s`; measured `2m41.2s` for 480 passes under
  `--check-bounds=auto`.
- **`GeometricMachineLearning.layer` resolves.** `src/GeometricMachineLearning.jl:88` has
  `import AbstractNeuralNetworks: layer`, and `AbstractNeuralNetworks/src/chain.jl:30` defines
  `layer(c::Chain, i) = c.layers[i]`. Indices 1–4 match the four `Chain` entries.
- **`test/reachability.jl` passes with the two entries gone** — `2/2`, and its own staleness check
  (an allowlist entry whose file is now reachable) is what would have caught a half-done removal.
- **`transformer_setup.jl` is wired in unchanged and asserts something real** — one `@test` on
  `ps[1].PQ.head_1 <: StiefelManifold`.
- **Formatting and encoding.** All three changed test files are already SciML-formatted
  (`format_text(src; style=SciMLStyle()) == src`). `nfc.jl` exits 0 on the changed files.
- **CHANGELOG entry present** under `## [Unreleased] — 0.8.0`, and the relabelling claim checks out:
  `sympnet_layers_test.jl` contains no symplecticity assertion.
- **Comment tense.** The new comment block at lines 3-7 and the three inline comments are all
  present tense and describe the file as it is. No historical notes outside the CHANGELOG.

### Not checked

- **Type piracy, inference, allocations.** The diff touches only `test/` and `CHANGELOG.md` — no
  `src/` change, so no new method, dispatch or inference surface. Aqua and JET were not run; nothing
  in this diff could move them.
- **The Float32-vs-Float64 failure-rate comparison** (12.7% vs 11.7%). Not re-measured. The stronger
  structural claim was verified instead: the round trip misses by 1e-2 to 5e-1, which no amount of
  precision closes.
- **Full `Pkg.test()`.** Not run locally; the CI matrix covers it.

michakraus:
Follow-up to [my review above](https://github.com/JuliaGNI/GeometricMachineLearning.jl/pull/275#pullrequestreview-5213195614). I pushed `f9bcb8c7`, which addresses the four **nit** items. The three **quality** items are untouched and still stand.

**Fixed**

- **#4** — dropped the fourth assertion from `test_symplecticity`. It checks column consistency, not symplecticity, and `layers/sympnet_layers_test.jl:31-33` already covers that for three layer chains. The now-unused `ten` binding goes with it.
- **#5** — the CHANGELOG said "max deviation 2–5e-7". Measured over 20 draws at each of the 120 swept pairs, the worst is **7.75e-7**, at `N=20, N2=80`. The entry now says `7.8e-7` and names the point.
- **#6** — the `see CHANGELOG.md` pointer now names the entry: `[Unreleased]`, `### Infrastructure`.
- **#7** — the deleted weight-tying comment's claim is now recorded in the CHANGELOG, with numbers. Tying is real but not sufficient: over 100 draws at each of three pairs it pulls the round-trip deviation down by an order of magnitude (median `3.4e-2` vs `4.7e-1` at `(2,4)`; `5.2e-2` vs `7.9e-1` at `(4,8)`; `3.3e-2` vs `7.7e-1` at `(10,20)`), and still exceeds `1e-5` in 299 of those 300 draws. The rank argument rules it out regardless of weights.

**Correction to my own finding #1**

My review claimed the 120-pair sweep is what costs `2m44.2s`, and that shrinking it would pay for itself. **That attribution is wrong.** Once Zygote has compiled the three Jacobian closures, the whole 120-pair sweep takes **0.3s**. The cost is one-time compilation, which does not scale with the pair count.

Timing the edited file cold: `360 passed, 2m49.1s` — removing a quarter of the assertions changed nothing measurable. So finding #1 should be read as "the sweep buys no information", which still holds, and **not** as "the sweep is expensive", which does not. The CHANGELOG now records this.

Re-running the sweep is therefore nearly free, and #1 is no longer a cost argument. I leave the decision on #1–#3 to you.

**Verification of the pushed commit**

- `test/layers/sympnet_upscaling.jl` — 360 assertions, 0 failures, cold process.
- `test/reachability.jl` — 2 passed.
- pre-commit: formatting ok, `fatou lint` clean, NFC ok, loads ok.

I have not merged this PR, and will not — I have now written code on the branch.

PR #278, opened 2026-09-16
Retire the train! subsystem, porting its content first

michakraus:
Phase B of the test- and script-tree restructuring: the `train!` subsystem is retired, and the
numerical content it held is ported onto `NetworkLoss` first so that nothing is lost with it.

Six commits, each self-contained and reviewable on its own:

| commit | what |
|:--|:--|
| `9c6ddbf1` | `LNNLoss` — a modern loss for `LagrangianNeuralNetwork` (T4) |
| `ac551faa` | `SymplecticEulerLoss`, `VariationalMidpointLoss` (T5) |
| `576ac38e` | delete `src/training/`, `src/nnsolution/`, `src/data/`, `src/training_method/` and the tests that exercised them (T6, T7) |
| `22c8f30c` | delete `scripts/test/`, modernise the pendulum scripts (T8) |
| `a6ddaaaf` | close `C8` (T19) |
| `1e3f0393` | answer the review — see *Review round* below |

## Why the ports come first

**Five of the six training methods were never executable.** Each line below was produced by running
the method, not by reading it.

- `LnnExactMethod` runs but ignores its own data. `loss_single` is
  `abs(sum(∇q∇q̇L(nn, qₙ, q̇ₙ)))`, and sweeping `q̈ₙ` over `-5`, `0`, `5` returns `0.291075` every
  time: the argument is accepted and never used. The Euler–Lagrange content is in the commented-out
  remainder of that same line, which is itself incomplete.
- `VariationalMidPointMethod` cannot run at all, for two independent reasons. `discrete_lagrangian`
  returns an array, so `Zygote.gradient` refuses it; and `DL₁`/`DL₂` slice a 2-tuple of gradients
  with ranges, which selects tuple entries rather than vector components — right by coincidence at
  one degree of freedom, wrong at every other.
- `SymplecticEulerA`, `SymplecticEulerB` and `HnnExactMethod` are **B6**: all three call
  `vectorfield`, which is defined nowhere.
- `BasicSympNetMethod` runs, at one degree of freedom only — `q̃ₙ₊₁, p̃ₙ₊₁ = nn([qₙ..., pₙ...])`
  destructures the output vector into two scalars, so 2 dof raises `DimensionMismatch`.

So the ports are repairs, not transcriptions.

## `LNNLoss`

Solves the Euler–Lagrange equations of the learned Lagrangian for the acceleration and compares it
against the data:

```
∇q̇∇q̇L q̈ + (∇q∇q̇L)' q̇ = ∇qL
```

The transpose is load-bearing, and the test discriminates on it: the solve is checked against a
closed form for `L = ½q̇'Mq̇ + q'Cq̇ - ½q'Kq` with `C` deliberately **not** symmetric, and the same
expression with the transpose dropped is asserted to *fail* that comparison.

Conditioning of `∇q̇∇q̇L` was measured before the solved form was chosen over an un-inverted
residual — median condition number `1.0`–`7.9`, worst `3332` over 600 draws — and is now asserted in
the suite rather than quoted from a run nobody can repeat.

## `SymplecticEulerLoss` and `VariationalMidpointLoss`

`SymplecticEulerLoss(arch, timestep)` trains an HNN from two consecutive states and a timestep, with
no velocities in the data. Variant `:A` evaluates the vector field at `(qₙ₊₁, pₙ)`, `:B` at
`(qₙ, pₙ₊₁)`.

`VariationalMidpointLoss(arch, timestep)` trains an LNN from three consecutive positions, through
the discrete Euler–Lagrange equations of the midpoint discrete Lagrangian.

**The manual has described the first of these since before it could be run.**
`docs/src/architectures/hamiltonian_neural_network.md` has a section *HNN Loss for Phase Space Data*
giving the symplectic-Euler-B formula, and the only implementations of it were `SEulerA`/`SEulerB`.

`BasicSympNetMethod` gets no port: measured on one network and one pair of states, it returns
`sum(abs2, prediction - target)` where `FeedForwardLoss` returns
`norm(prediction - target) / norm(target)` — the same residual, unnormalised and squared.

## Nested `Zygote` is not available, so the derivatives are symbolic

Differentiating a loss that itself calls `Zygote.gradient` with respect to the parameters fails with
`MethodError: no method matching getindex(::IdDict{Any, Any})`, measured in two separate processes.
Both new LNN losses therefore take their input derivatives from compiled symbolic expressions, the
route `hamiltonian_vector_field` already took. For the variational loss this meant writing the chain
rule through the midpoint out by hand, so that only the network's *first* input derivative is needed;
those two expressions are checked against a `Zygote` gradient of `L_d` itself, which is the call they
exist in order not to make.

## What the retirement removes

`src/training/` (7 files), `src/nnsolution/` (3), `src/data/` (5), `src/training_method/` (7) and
`src/architectures/default_architecture.jl` — 1,438 lines — plus 820 lines of tests.

**The package exports 130 names where it exported 218.** Measured from
`names(GeometricMachineLearning)` before and after, not from the diff, which over-reports by 88
because several names are exported on more than one line and only one line was removed. `dim`,
`iterate`, `solve!`, `apply!`, `jacobian!` and `description` were each checked to have survived.

**Six of the deleted test files were passing.** `test/data/test_data.jl`, `test/data/test_batch.jl`,
`test/data/test_matching.jl` and `test/training_parameters.jl` ran on every suite and covered
`TrainingData` construction, the old `get_batch`, `matching`, and `TrainingParameters`. All of it
tested code that no longer exists, so none of it could be kept — but the subsystem leaves with its
tests rather than its tests having been absent.

## Both guards' allowlists are now empty

`test/exports.jl` passes with `UNDEFINED_EXPORTS` empty and `test/reachability.jl` with
`ALLOWED_ORPHANS` empty, each run directly. Every exported name resolves and every file under
`test/` is reached from `runtests.jl`.

That closes **C10**: seven exports for names defined nowhere are dropped (`Device`, `CPUDevice`,
`convert_to_dev`, `ResidualLayer`, `LinearSymplecticLayerP`, `LinearSymplecticLayerQ`, and `timestep`
with its block), and `description` gains the `import GeometricBase: description` its own export
comment implied.

## The scripts run

- `scripts/hnn_pendulum.jl` — **this is the script B6 was diagnosed in**, and this repository
  recorded that it "still does not run to completion, and what stops it is this". It now trains on
  `HNNLoss`: loss `1.0017 → 0.0041` over 2000 epochs, plot written, exit 0.
- `scripts/Script_using_fully_GML/lnn_script.jl` — had never run under any API generation. It
  referenced `TrainingIntegrator` and `DataTrajectory`, neither of which exists, and included a path
  that does not exist. It now trains on `LNNLoss`: `3.248 → 0.175` over 200 epochs, exit 0.

`grep -rn 'TrainingData\|train!' scripts/` returns nothing.

`LNNLoss` needs a larger step size than `HNNLoss` — `1e-2` against `1e-3` — because its gradient
reaches the parameters through a solve against the velocity Hessian, where `HNNLoss` is linear in the
gradient of the learned Hamiltonian.

## Issues closed

**B6**, **C8**, **C10** and **C11**, each removed from *Open Issues* rather than annotated, per that
section's own rule. **B5**, **C1**, **C2**, **C6**, **C7**, **C9**, **C12** and **D4** were each
checked against the tree and still describe the current state.

## Scope notes

Two sub-task boundaries moved, both forced:

- **T4 and T5 are stacked rather than parallel.** Both need a Lagrangian documentation page —
  `docs/make.jl` sets neither `checkdocs` nor `warnonly`, so Documenter's default `:all` makes a
  docstring in no `@docs` block a new `missing_docs` error class. Two independent branches would each
  have created that page and conflicted.
- **T6 and T7 are one commit.** Deleting `src/data/` while `test/data/test_data.jl` still exercises
  it cannot produce a green pull request.

Three things were found and deliberately **not** fixed here:

- `Aqua` is in neither `Project.toml` nor `test/`, so the 13 type-piracy hits it reports are first
  observations with no baseline. None is in this diff.
- The `Documentation` job is already red at the merge base, with an error set this branch does not
  change. The merge-base run `35010496235` at `efc5235b` and this branch's run `35065961657` both
  carry 62 `invalid local link/image`, 2 `Cannot resolve @ref` and one `@example` `MethodError` in
  `docs/src/tutorials/optimizer_comparison.md`, and both terminate on
  `[:cross_references, :example_block]`. Neither adds a `missing_docs` class, so the new pages and
  docstrings cost the job nothing.
- `apply!` and `jacobian!` remain method-less generics in `src/backends/backends.jl`. They were dead
  before this work.

## Review round

`1e3f0393` answers a review of this branch. Four changes, none in the numerics:

- `CHANGELOG.md` contradicted itself. Two sentences under *Added* said the `train!` subsystem and
  `BasicSympNetMethod` were "still present and still exported", which held for `ac551faa` alone and
  not for the branch, where *Removed (breaking)* deletes them. A changelog is read as one release.
- `∇q∇q̇L` at `src/architectures/lagrangian_neural_network.jl:43` was orphaned by this branch — its
  one caller was the deleted `src/training_method/lnn_exact_method.jl:10` — and duplicated what
  `lagrangian_acceleration` builds symbolically. It is deleted. Its three dead neighbours predate
  this work and are recorded as **C13** instead.
- `Distances` is dropped from `[deps]` and `[compat]`. The entry was going to outlive the `using`
  that this branch removed, in a release that is already breaking.
- `docs/src/architectures/hamiltonian_neural_network.md` now warns that the displayed formula's
  second term disagrees in sign with `SymplecticEulerLoss`. The sign predates the implementation,
  but the page had never cited an implementation before.

Phase E — restructuring `test/` and `scripts/`, and adding the CI job that gates `scripts/` — is not
in this pull request.


michakraus (commented):
## Review

The code and the mathematics are correct. I checked the Euler–Lagrange algebra, the transpose
argument and the midpoint chain rule by hand, and each holds. All eight Julia CI jobs pass,
`windows-latest` and both `min` jobs included. Both guards are non-vacuous with their allowlists
empty: `test/exports.jl` keeps its staleness assertion and `test/reachability.jl` keeps its closure.
No deleted name survives anywhere in `src/`, `test/`, `docs/src/` or `scripts/`.

Five findings. None is in the numerics. Four are prose or orphaned code, and I push the corrections
for all of them in a follow-up commit.

### What I verified by hand

| claim | result |
|:--|:--|
| `∇q̇∇q̇L q̈ + (∇q∇q̇L)' q̇ = ∇qL` | correct |
| the transpose is load-bearing | correct. `H[1:n, (n+1):2n][i,j] = ∂²L/∂qᵢ∂q̇ⱼ` is the off-diagonal block and is not symmetric, so `M'q̇ ≠ Mq̇` |
| the closed form in the test | correct. `H = [-K C; C' M]` and `g = [Cq̇ - Kq; Mq̇ + C'q]` give `q̈ = M\((C - C')q̇ - Kq)` |
| the transpose-dropped comparison discriminates | correct. It collapses to `M\(-Kq)`, which differs whenever `(C - C')q̇ ≠ 0` |
| `D₁Lᵈ = (Δt/2)∇qL - ∇q̇L`, `D₂Lᵈ = (Δt/2)∇qL + ∇q̇L` | correct |
| the DEL residual assembly `D₂Lᵈ(qₙ, qₙ₊₁) + D₁Lᵈ(qₙ₊₁, qₙ₊₂)` | correct |
| `:A` at `(qₙ₊₁, pₙ)`, `:B` at `(qₙ, pₙ₊₁)` | correct, and `:B` matches the integrator the manual names |
| `dim(arch) ÷ 2` | correct. `dim(::LagrangianNeuralNetwork) = arch.dimin`, the stacked input dimension |
| the scripts run | corroborated. Both plots sit in the author's worktree, written today at 07:19 and 07:21 |
| commit subjects | all five are 50 characters or fewer |

### F1 — `CHANGELOG.md` contradicts itself

`CHANGELOG.md:643` says the `train!` subsystem "and its methods are still present and still
exported". `CHANGELOG.md:664` says `BasicSympNetMethod` "is still defined and still exported — this
change removes no file".

Both sentences were true for `ac551faa` on its own. Both are false at the branch tip, where
`### Removed (breaking)` in the same unreleased version deletes that subsystem. A reader reads the
changelog as one release, not as a commit sequence.

### F2 — the deletion orphans `∇q∇q̇L`

`src/architectures/lagrangian_neural_network.jl:43`. Its only caller was
`src/training_method/lnn_exact_method.jl:10`, which this branch deletes. I checked that against
`origin/main`, so the attribution is to this diff and not to an earlier one.

It is now a second implementation of a quantity `lagrangian_acceleration` builds symbolically. That
is the argument this pull request already makes for `data_problem.jl`: one copy of a formula beside
another copy is worse than either alone.

Three neighbours in that file are dead too — `∇L:29`, `∇∇L:35`, `∇q̇∇q̇L:39` and the constant
`DEFAULT_LNN_NRUNS:1`. They were dead before this branch: `∇L` had no caller at all, and
`∇q̇∇q̇L` appeared only in the commented-out remainder of that same line 10. `∇∇L` keeps a caller
in `∇q̇∇q̇L`, so this branch does not orphan it either. Those four are reported, not deleted.

### F3 — `Distances` is now a dead dependency

No file under `src/`, `test/`, `docs/src/` or `ext/` uses it. The description discloses this and
defers the two `Project.toml` lines to "a change that owns that file". This release is already
breaking, and it already owns the `using` line that made the dependency dead. A later change has no
reason to touch it, so the entry keeps shipping to every user.

### F4 — the description under-counts the `Documentation` failures

The claim itself holds, and I checked it against both logs. The merge-base run `35010496235` at
`efc5235b` and this branch's run `35065961657` carry a byte-identical error set and the same
terminating categories, `[:cross_references, :example_block]`. Nothing here reddens a green job.

The count is short by one class. Both runs also carry 2 `Cannot resolve @ref`, beside the 62
`invalid local link/image` and the one `@example` `MethodError` in
`docs/src/tutorials/optimizer_comparison.md`.

### F5 — the HNN manual page now cites code that disagrees with the formula above it

`docs/src/architectures/hamiltonian_neural_network.md:39` points the reader at
`SymplecticEulerLoss` and states the convention `X_H = (∂H/∂p, −∂H/∂q)`. The displayed formula's
second term reads `+ (q^{t+1} − q^t)/Δt`, where that convention gives a minus.

The sign is pre-existing and this branch is right to leave it. What changed is that the page named
no implementation before, so the disagreement was invisible. It is now a cross-reference from a
formula to code that does not compute it.

### Notes, no action asked

- `lagrangian_acceleration` solves one `n × n` system per sample inside a comprehension, and Zygote
  differentiates through it. It is the clear implementation, and it is why the test file records
  61 s for a fresh `Chain` specialisation.
- The `SymplecticEulerLoss` constructor takes its `:A`/`:B` type parameter from a runtime `Symbol`,
  so construction does not infer. It runs once per training, so this costs nothing.
- Both new manual pages put underscore-prefixed functions in `@docs` blocks. `checkdocs` defaults
  to `:all` here, so that is forced rather than chosen.

**Verdict: approve, subject to F1.** F1 ships a contradiction in the release notes and has to go
before this merges. F2, F3 and F5 are small and I push them with it. F4 is the description only.

I did not merge this and will not: I am not its author, but I now author part of it.

michakraus:
## Review fixes pushed — `1e3f0393`

All five findings are answered. Four in the branch, one in the description.

| finding | what changed |
|:--|:--|
| F1 | `CHANGELOG.md` — the two sentences that said the `train!` subsystem and `BasicSympNetMethod` were "still present and still exported" now describe the release, not the commit |
| F2 | `∇q∇q̇L` deleted from `src/architectures/lagrangian_neural_network.jl` |
| F3 | `Distances` dropped from `[deps]` and `[compat]` |
| F5 | a `!!! warning` under the formula in `docs/src/architectures/hamiltonian_neural_network.md` |
| F4 | the description now names both runs and all three error classes |

The three dead neighbours of `∇q∇q̇L` predate this branch, so they are recorded as a new **C13**
under *Open Issues* instead of deleted. The changelog entries for the two code changes are in the
same commit that earns them.

`Pkg.test()` on Julia 1.11: exit 0, `Testing GeometricMachineLearning tests passed`, no failures,
2 pre-existing `@test_broken`. `Lagrangian Neural Network` passes 34 and `Symplectic Euler and
variational midpoint losses` passes 36. Both guards still hold with their allowlists empty:
`Reachability of every file under test/` passes 2 and `Exported names are defined` passes 2.
`Project.toml` came back from the run byte-identical, with no `[sources]` block injected.

### Two corrections to my own review

- I wrote that `∇∇L` "loses its last caller too". It does not. `∇q̇∇q̇L` still calls it, so this
  branch does not orphan it, and it is not deleted. Only `∇q∇q̇L` is.
- I cited `∇L:29` and `DEFAULT_LNN_NRUNS:1`. The correct lines are `30` and `2`; line 1 is blank.
  C13 carries the right ones.

I also nearly wrote into C13 that `∇L` never had a caller. `git log -S '∇L(nn'` shows it had one
until `7f1b3dd0` (2023-06-22), so the entry says that instead.

### Not merging this

I reviewed it and I now author part of it. CI has to go green on `1e3f0393` first, and the merge is
not mine to make.

PR #279, opened 2026-09-16
Restructure test/ by subject, add test/Project.toml

michakraus:
Phase E, T17 of `Tasks/Restructure the GeometricMachineLearning test and script trees.md`.

`test/` becomes one directory per subject — twelve of them, none holding a single file — and each
carries its own driver naming the testsets it runs. `test/runtests.jl` is twelve `include` lines
and the two guards, where it was 62 inline `@safetestset` blocks. `test/Project.toml` replaces the
`[extras]`/`[targets]` pair in the package's own `Project.toml`.

## The suite runs exactly what it ran before

62 testsets, 4085 passes, 2 `@test_broken`, 0 failed — on `origin/main` and on this branch, with
the same testset names and the same per-testset pass counts, compared row by row rather than by
grand total. The names lose their trailing space padding, so the `Test` summary columns are
narrower; nothing else about them changes.

The six `@info "Starting …"` markers go with the inline blocks. Each `@safetestset` already prints
its own summary as it finishes.

## Why the drivers are included at top level

`@safetestset` expands to a `module`, and a module may not appear inside a testset body — so
`include("layers/runtests.jl")` at top level, not `@testset "Layers" begin … end`. The two guards
stay beside `runtests.jl`: they check the tree rather than a subject in it, and
`test/reachability.jl` reads `test/` off its own `@__DIR__`.

## Pre-PR verification

Run by `julia-branch-verifier` against the diff, then re-derived here.

**Fixed after it reported**
- Five claims in the `CHANGELOG.md` entry were wrong and are corrected: four single-file
  directories rather than six, one loose attention test rather than two, 35 renames rather than 36,
  testset names identical only after their padding is dropped, and a wrong mechanism for what
  `test/Project.toml` buys.
- `test/parameters/network_parameters_gradient_structure.jl:24` and
  `test/reduced_order_modeling/sae_error_lower_than_psd_error.jl:9` — comments naming test paths
  this change moved.

**Checked and clean**
- Renames: 35 `R` entries, 33 at 100% similarity, no delete/add pair. Four files carry a content
  change: two an `include` path, two the comments above.
- `test/Project.toml`: the 16 `[deps]` are exactly the packages `using`/`import`ed under `test/`,
  derived independently from the files. `[compat]` covers 15 of 16, and the six names shared with
  the package's own `Project.toml` carry identical bounds.
- The package's `[compat]` no longer names a package it does not depend on.
- Formatting (`style = "sciml"`), NFC and `fatou lint` clean on every changed `.jl` file.
- No stale reference to any of the 62 old test paths survives outside `CHANGELOG.md`.

**Pre-existing, reported not fixed**
- `test/runtests.jl:1` imports `Test` and `GeometricMachineLearning`; neither is used.
- `test/parameters/changebackend_tests.jl:5` keeps history in a comment, and its path is stale.
- `Project.toml` has no `[compat]` for the stdlib deps `InteractiveUtils`, `LinearAlgebra`,
  `Random`.

**Not checked**: type piracy, inference, allocations and doctests — the diff changes no `src/` line
and no docstring.


michakraus:
`Documentation` is red, and it is red on `main` too — the last three runs on `main` (`3c7d6ec6`, `424d3f0c`, `efc5235b`) all failed. This branch changes no file under `docs/` and no docstring in `src/`, and `Doctests - ubuntu-latest` passes here. The other eleven checks are green: all nine matrix jobs, doctests, and both codecov reports.

michakraus (commented):
Reviewed against a clean worktree of the branch. The restructure is sound, and I could re-derive
the central claim independently rather than take it from the description. I found three wrong
statements, all one-liners and none of them behavioural, and I have pushed the corrections as
`1457436a` rather than leave them for you.

## What I re-derived rather than took on trust

**The suite runs exactly what it ran before — confirmed three ways.**

1. `main` CI versus branch CI, same job (`Julia 1 - ubuntu-latest`), one parser over both logs:
   the testset name lists and the per-row numbers are equal, name for name and number for number.
2. `main` CI versus a local `Pkg.test("GeometricMachineLearning")` on the branch (exit 0): after
   the name padding and the time column are dropped, the two lists of 62 `Test Summary` blocks
   **diff to nothing**. Totals 62 testsets, 4085 passes, 2 `@test_broken` on both sides.

**No file dropped, no orphan, no double-run.** An AST walk of every `runtests.jl` on each ref gives
62 `@safetestset` names mapping to 62 unique files on both, name for name to the same basenames.
The one `.jl` outside that set on **both** refs is the shared helper
`network_parameters_gradient_structure.jl`, which two tests `include` directly.

`test/reachability.jl` still works under the new layout — run standalone on the branch it reports
`2 2`. Its walk recurses into the `@safetestset` macro call, so the drivers' includes are inside
the closure.

`test/Project.toml`: a sweep of every `using`/`import` under `test/` yields exactly the 16 declared
`[deps]` and nothing else. `[compat]` covers 15, the gap being the package itself. The six names
shared with the package `Project.toml` carry byte-identical bounds. The package `[compat]` has no
name outside `[deps] ∪ [weakdeps]`, `[deps]` is untouched, and nothing under `test/` reaches an
undeclared module. `julia --project=test -e 'using ForwardDiff'` does fail, as the entry says.

`.githooks/pre-push` takes the `Pkg.test` branch, so it honours `test/Project.toml`. Formatting
(`style = "sciml"`), NFC and `fatou lint` are clean on all 13 drivers.

## Three corrections, now pushed as `1457436a`

1. **`test/parameters/changebackend_tests.jl:5` named a path this branch moved.** The comment read
   "These testsets used to live in `test/hdf5_support.jl`"; the file is now
   `test/parameters/hdf5_support.jl`. The description files this under *pre-existing*, but it was
   correct on `main` — this branch is what made it dangle, and it is the same class as the two
   comments that were fixed. It is also the counter-example to "No stale reference to any of the 62
   old test paths survives outside `CHANGELOG.md`": I grepped all 35 removed paths and the nine
   removed directory names over `src/`, `docs/`, `scripts/`, `.github/`, `.buildkite/` and
   `.githooks/`, and this was the only hit.

2. **`CHANGELOG.md` — "The six `@info "Starting …"` markers" is seven.**
   `git show origin/main:test/runtests.jl | grep -c '^@info'` gives 7, at lines 15, 71, 103, 134,
   148, 153 and 190. `CHANGELOG.md:1877` already records "emits **seven** `@info` markers".

3. **`CHANGELOG.md` — "35 renames, 33 of them at 100% similarity" was 31, and is 30 after fix 1.**
   The entry contradicted itself two sentences later: "Four files carry a content change", and
   35 − 4 = 31. Correcting the comment in fix 1 makes that file a fifth content change, so the
   entry now reads 35 renames, **30** at 100% similarity, **five** files with a content change, and
   lists `parameters/changebackend_tests.jl:5` beside the other two comment fixes.

Re-derived after the commit: `grep -c '^R100'` → 30, `grep -c '^R'` → 35, and the five non-`R100`
entries are exactly the five the entry names. The pre-commit hook reports formatting, `fatou lint`,
NFC and load all clean.

## CI

On the previous head `7d6477f6` the whole matrix was green: the seven required contexts — the six
`Julia {min,1} - {ubuntu,macOS,windows}` jobs and `Doctests` — plus `Julia pre` and `Julia nightly`.
`1457436a` changes one comment line and one CHANGELOG paragraph, so CI is re-running on it now.

`Documentation` fails, but **it fails identically on `main`** — the same first error,
`failed to run @example block in docs/src/tutorials/optimizer_comparison.md:3-36`, then the
`tikz/*.png` `invalid local link/image` set, ending in
`makedocs encountered errors [:cross_references, :example_block]`. The last three `Documenter.yml`
runs on `main` are all `failure`. Not caused here, and not a required check.

## Noted, not fixed

- `test/runtests.jl:1` still imports `Test` and `GeometricMachineLearning` unused. Line 1 is
  unchanged context, so it is genuinely pre-existing — but the rest of the file is rewritten here,
  which makes it a natural time to drop them.
- The subject order changes the order the suite runs in, and ten randomised test files never call
  `Random.seed!` (`kernels/tensor_inverse.jl`, `attention/test_skew_map.jl`,
  `attention/test_cayley_transforms.jl`, `layers/sympnet_upscaling.jl`,
  `layers/volume_preserving_feedforward.jl`, `optimizers/gml_gradient_dispatch.jl`,
  `arrays/triangular.jl`, and three under `docstrings/`). Pre-existing, and the identical
  per-testset counts say nothing moved — worth knowing the exposure exists.

**Verdict: approve once CI is green on `1457436a`.** The restructure provably runs the same 62
testsets with the same counts, the reachability guard survives the layout, and the test environment
is now exactly what it declares. I authored the correction commit, so the merge is not mine to make.

PR #281, opened 2026-09-17
Write the HNN tutorial's phase space section

michakraus:
## What

`docs/src/tutorials/hamiltonian_neural_network.md` ended at the bare heading *Training a HNN Based on Phase Space Data*. The page named a second way of training a Hamiltonian neural network and then stopped.

`SymplecticEulerLoss` is implemented in `src/loss/symplectic_euler_loss.jl`, documented in `docs/src/architectures/hamiltonian_neural_network.md` and tested in `test/losses/training_method_losses.jl`. Only the tutorial's promise of it was empty.

## How

The new section trains the **same** Hamiltonian as the first one, so the two differ only in the format of the data:

| section | data | loss |
|:--|:--|:--|
| Vector field data | the field ``𝕁z`` on a grid of the domain | `HNNLoss` |
| Phase space data | the pairs that grid is carried to by the exact flow ``exp(Δt𝕁)`` | `SymplecticEulerLoss` |

It continues the existing `@example hnn` block, so it reuses `domain_matrix`, `hnn_arch`, `batch` and `n_epochs` rather than rebuilding them, and it starts from a fresh network.

It also names the variant the two-argument constructor gives, which the architecture page leaves to the docstring: that page's formula is written for `:B`, and the constructor defaults to `:A`.

## The page now seeds its randomness

`import Random` and `Random.seed!(1234)`, at the top of the first of the page's eight `@example` blocks, in the form `docs/src/tutorials/sympnet_tutorial.md:26-27` already uses.

The network initialisation was a fresh draw on every build, so both loss plots changed from one build of the manual to the next and no figure on the page could be quoted. **This changes the first section's vector-field plot as well as the new one** — that is the one place this PR alters output the page already ships.

## Why `@example` and not a fenced block

A plain ```` ```julia ```` fence is rendered and never executed, which is the defect class this repository is removing. The added blocks are `@example`, and they cost less than what the page already pays:

| blocks | build time |
|:--|--:|
| the four already on the page | 77 s |
| the four added here | 40 s |

**Read those two figures as build cost, not as a comparison of the two losses.** The added blocks run second, so they do not pay the compilation the first four have already paid. Of the added 40 s, 30 s is the training and 8.75 s is building the symbolic vector field in the `SymplecticEulerLoss` constructor. A second run of the same script gave 42 s against 80 s — the ratio is stable, the absolute seconds are not.

Timed like for like instead, one fresh process per loss, each process training twice so that run 1 pays compilation and run 2 does not:

| loss | cold | warm |
|:--|--:|--:|
| `SymplecticEulerLoss` | 60.49 s (59.28 s in a second process) | 25.98 s |
| `HNNLoss` | 62.12 s | 27.47 s |

The two losses cost the same to train against within a few percent. **Treat every second in this PR as an upper bound:** two `Pkg.test()` runs from another session were active across the measurement window, so no run had the machine to itself. The conclusion does not turn on the precision — a 2.7 % gap is inside single-run noise either way, and what justifies `@example` is that the added blocks cost less than the page already pays.

## Verification

- **Every `@example hnn` block of the page evaluated in order, in one fresh module** — which is what Documenter does — in the docs environment. Exit 0. Per-block seconds: 2.82, 0.35, 5.73, 68.42 for the pre-existing four, and 0.60, 0.00, 8.75, 30.34 for the added four.
- **The seed reproduces the page.** Two independent cold processes gave bit-identical curves: `(1.1469659392288465, 0.027395355229781104)` for the first section and `(1.3568450815737088, 0.015629138834060172)` for the added one. A seed that left the batch shuffling free would still have drifted, so this is the check that the seed governs the whole block sequence and not just the initialisation. Before the seed, three runs gave first values of 0.94, 1.09 and 1.63.
- **Both curves fall**, so the hidden `lines` calls plot something. Both losses are relative residuals, which is why both start near 1.
- **`exp(Δt * Matrix(𝕁))` is the exact flow of `vf`.** It printed `[0.99500 0.09983; -0.09983 0.99500]`, i.e. the rotation by `Δt = 0.1`, as it must for `vf(z) = 𝕁 * z`.
- **`docs/check_references.jl`: 0 unresolved references, exit 0.** This is the cheap half of the Documentation job — `@docs` resolution and code `@ref` resolution. It does not cover `@example` blocks, which the run above covers instead, nor section-title `@ref`s; the one added section-title `@ref` names a header that exists verbatim at `docs/src/architectures/hamiltonian_neural_network.md:23`.
- **NFC: both changed files clean, exit 0.** `Δt` and `𝕁` are single codepoints in NFC. The checker was shown to exit 1 on a deliberately NFD file, so the clean result is not a silent pass.

Worth knowing when reading the commits: the shared `pre-commit` hook filters staged files to `*.jl` and exits 0 when none match. **A Markdown-only commit runs no formatter, no NFC stage and no load test**, so every check above is a deliberate run, not the hook's.

## Pre-PR verification

Run in its own context against `main...HEAD`. **Verdict: clear, nothing unresolved.**

Fixed during the pass:

- `docs/src/tutorials/hamiltonian_neural_network.md:57` — "the data are **no longer** vector fields" → "are **not** vector fields". Docs describe the current state; the CHANGELOG carries the transition.
- The timing claim in the CHANGELOG and in this body. "28 s against 61 s, timed on their own in a cold process" set a steady-state figure against a cold one, so it measured whether compilation had been paid, not the two losses. There is no factor of two between them.

Verified from the source rather than from printed output:

- `Matrix(𝕁)` is the dense `[0 1; -1 0]`, which agrees with `𝕁 * [q; p] = [p; -q]` at `src/arrays/poisson_tensor.jl:70`. `vf` is linear, so the time-`Δt` flow **is** that matrix exponential.
- `SymplecticEulerLoss(arch, timestep)` defaults to `:A` at `src/loss/symplectic_euler_loss.jl:47`, and both variants' evaluation points match the prose (lines 64-74).
- The timestep enters the loss only through `difference_quotient`, line 87, so "the loss needs the timestep" is exact.
- Every name the new blocks bind is new; `batch` and `n_epochs` are read and never reassigned, so no earlier block is disturbed.
- The section-title `@ref` was checked by hand, since `check_references.jl` states it does not cover those. The target title occurs exactly once in `docs/src`.

Pre-existing and adjacent, **left alone** — raising them here rather than fixing them in a docs PR:

- `docs/src/architectures/hamiltonian_neural_network.md:42` — a `!!! warning` narrates history in a docs page: the sign "predates the implementation". This tutorial now links to that section twice, so the diff raises its traffic.
- Four duplicated header titles in `docs/src`: `Library Functions`, `References`, `Architectures`, `Loss function`. A section-title `@ref` to any of them would be ambiguous and fail the build. None is referenced today.

## Review round

Reviewed at `40ed5cf1`. The review is in the thread; every cheap source-level claim in this body reproduced. Three findings and two nits were raised, and all but one are now fixed on this branch.

**The page did not say what it trains towards.** The data comes from the exact flow, but the loss holds the network to one symplectic Euler step. The two agree only to first order in `Δt`, so the minimiser is not the Hamiltonian of the system, ``H(q,p) = (q^2+p^2)/2``, but the Hamiltonian whose symplectic Euler step reproduces that flow. Without saying so, a trained loss **below** the true Hamiltonian's own residual reads as a fit better than exact. The page now carries an `!!! info` that states it.

`scripts/verification/symplectic_euler_loss_modified_hamiltonian.jl` establishes the claim, in the directory whose stated purpose is "checks that establish a mathematical claim about the package", so the scripts gate runs it from the moment it lands:

| Δt | residual of the true `H`, variant `:A` | variant `:B` |
|--:|--:|--:|
| 0.1 | 0.05000694627749635 | 0.05000694627749525 |
| 0.05 | 0.025000868112826917 | 0.025000868112826924 |
| 0.025 | 0.012500108508734082 | 0.012500108508734085 |

Halving `Δt` halves the residual — first order, as the note claims. The script asserts rather than prints, and the assertion was shown to fire: on data generated by the symplectic Euler step itself, rather than by the exact flow, the residual is `4.74e-16`.

**Two smaller fixes on the page.** `QPT`, `norm` and `gradient` were imported in the first block and used nowhere — this body listed them as left alone, but the seed is inserted directly beneath them, so they are inside the block this PR edits. They are gone. And the link to the architecture page's loss formula now says that the formula states the `:B` evaluation points and that the warning beside it holds, rather than sending the reader to a formula that page disowns.

**Two changelog nits.** "the page's single `@example` block" — the page has eight, and the seed is in the first; the same wording in this body is corrected above. And the changelog recorded five timing figures and disclaimed them in the same sentence. A changelog is not corrected later, so the seconds are gone from it; they remain here, which is where a measurement taken on a busy machine belongs. Both conclusions they supported stand: the added blocks cost about half of what the page already pays, and the two losses cost the same within a few percent.

**Not fixed, deliberately.** The seed is `# hide`, so a reader who copies the rendered page still gets different numbers from the figures. That matches `sympnet_tutorial.md:26-27`, so it is the house form rather than a defect.

## The `Documentation` job is red on `main` already

Do not read a red `Documentation` check here as this branch's doing. **The baseline is exact:** this branch's merge base is `61fa7fe8`, which is `main`'s tip, and that commit's own `Documentation` run ([35187942641](https://github.com/JuliaGNI/GeometricMachineLearning.jl/actions/runs/35187942641)) failed with **66** `##[error]` lines, on two causes that predate this branch:

- a `MethodError` in an `@example` block of `docs/src/tutorials/optimizer_comparison.md:3-36` — `Optimizer(::GradientMethod, ::@NamedTuple{xy::Matrix{Float64}}; step_size)`;
- a long list of `invalid local link/image` errors across ten or more pages, none of them this one.

A `grep -icE "sympnet_tutorial|hamiltonian_neural_network"` over that log returns **0**, so this page contributes nothing to main's red. Six consecutive runs on `main` are red.

The test for this branch's run is therefore: **any error naming `tutorials/hamiltonian_neural_network.md`, or a count above 66, is this branch's.**


michakraus (commented):
## Review

Reviewed `40ed5cf1cba332922c9b9bcec68013882284d6e1`, the tip of `refs/pull/281/head`. The merge base is `61fa7fe8`, which is `origin/main`'s tip, so the branch is not behind.

**Verdict: approve, conditional on the `Documentation` job.** The content is correct and the mathematics holds under independent computation. Nothing below blocks the merge. I am pushing fixes for findings 1 to 5 to this branch; finding 6 is the house form and stays.

### Findings

| # | severity | location | finding |
|--:|:--|:--|:--|
| 1 | quality | `docs/src/tutorials/hamiltonian_neural_network.md:59-63` | The data comes from the **exact** flow, but the loss enforces one **symplectic Euler** step. The two differ at O(Δt), so the network fits a modified Hamiltonian rather than the true one. The page does not say so, and the first section does not have this gap. Evidence: the variant `:A` relative residual of the true `H = (q² + p²)/2` on this exact data is `0.05000694627749635` (`:B`: `0.0500069462774952`). The PR reports the trained network at 0.016, so it beats the true Hamiltonian threefold. That is expected for a modified-equation fit, and unexplained on the page. |
| 2 | quality | `docs/src/tutorials/hamiltonian_neural_network.md:11-13` | `QPT`, `norm` and `gradient` are imported, rendered to the reader, and used nowhere on the page. The PR inserts lines 14-15 directly beneath them, so this is inside the block the PR edits. `QPT` occurs only at line 11, `norm` only at line 12, `gradient` only at line 13; the word at line 55 is prose. |
| 3 | quality | `docs/src/tutorials/hamiltonian_neural_network.md:82` | "which is the variant the [loss formula](@ref "HNN Loss for Phase Space Data") is written for" sends the reader to a formula the target page itself disowns. `docs/src/architectures/hamiltonian_neural_network.md:41-42` carries a `!!! warning`: "The displayed formula is not what that loss computes… the implementation is the one to read." The statement about the evaluation point is correct. The link needs the signpost. |
| 4 | nit | `CHANGELOG.md:985` | "the page's **single** `@example` block". The page has eight, at lines 9, 29, 38, 45, 61, 70, 77 and 86. The seed sits at the top of the first. The same wording is in this PR body. |
| 5 | nit | `CHANGELOG.md:972-979` | The changelog records five timing figures and disclaims them in the same paragraph ("Treat every figure here as an upper bound: no run had the machine to itself"). A changelog is not corrected later, so a reader in a year reads 40 s, 77 s, 60 s and 26 s as fact. The two qualitative conclusions survive without the numbers, and this PR body is the right place for the measurements. |
| 6 | nit | `docs/src/tutorials/hamiltonian_neural_network.md:14-15` | The seed is `# hide`, so a reader who copies the rendered page still gets different numbers from the figures. The stated motivation is that no figure could be quoted, which this meets only for the built page. This matches `docs/src/tutorials/sympnet_tutorial.md:26-27`, so it is the house form and not a defect. No change. |

### PR-body claims checked

Every cheap source-level claim reproduced.

| claim | held |
|:--|:--|
| `SymplecticEulerLoss(arch, timestep)` defaults to `:A` at `src/loss/symplectic_euler_loss.jl:47` | yes — `variant::Symbol = :A` |
| The prose about the `:A` and `:B` evaluation points matches the implementation | yes — lines 64-68 give `:A` as (qₙ₊₁, pₙ), lines 70-74 give `:B` as (qₙ, pₙ₊₁) |
| The timestep enters the loss only through `difference_quotient` | yes — line 87, and nowhere else |
| Every name the new blocks bind is new, and no earlier block is disturbed | yes — first occurrence of `Δt`, `next_matrix`, `dl_pairs`, `hnn_pairs`, `loss_pairs`, `o_pairs` and `loss_array_pairs` is in the added blocks; `batch`, `n_epochs`, `hnn_arch`, `domain_matrix` and `𝕁` are read and never reassigned |
| The section-title `@ref` target exists verbatim and is unique in `docs/src` | yes — `## HNN Loss for Phase Space Data` at `docs/src/architectures/hamiltonian_neural_network.md:23`, the only header with that text |
| `exp(Δt·𝕁)` is the exact flow of `vf(z) = 𝕁z` | yes — `Base.:*(::PoissonTensor, v::QPT)` at `src/arrays/poisson_tensor.jl:70` agrees with `[0 1; -1 0]`, and `vf` is linear. Computed independently: `exp(0.1·[0 1; -1 0]) = [0.9950041652780257 0.09983341664682814; -0.09983341664682814 0.9950041652780258]`, the value this PR prints |
| Both changed files are NFC clean | yes — `nfc.jl` exits 0 on both, and the script signals a fault with exit 1 |
| The seed uses the form `sympnet_tutorial.md:26-27` already uses | yes — those lines are exactly that form |
| Four duplicated header titles in `docs/src` | yes — `Architectures`, `Library Functions`, `Loss function`, `References`. None is a new `@ref` target |
| The per-block seconds sum to the quoted totals | yes, as arithmetic — 77.32 and 39.69, a ratio of 0.513. Not re-measured |

One claim this PR body does not make, checked because an unexported name would be an `UndefVarError` in the build: `SymplecticEulerLoss` is exported at `src/GeometricMachineLearning.jl:259`, so the unqualified call at tutorial line 78 resolves.

### Not checked

- The loss values 1.36 → 0.016 and 1.15 → 0.027, and the two-process reproducibility. Both need the training run.
- The absolute timings. Only their arithmetic was confirmed.
- `docs/check_references.jl`. The docs environment needs a resolve, and `BrenierTwoFluid` is a `rev = "main"` source, so running it risks rewriting `docs/Manifest.toml`. Its marginal value here is one code `@ref`, `GeometricMachineLearning.SymplecticEulerLoss`, which was verified by hand against the `@docs` block at `docs/src/architectures/hamiltonian_neural_network.md:54`.

### CI

The baseline for the `Documentation` job is exact. This PR's merge base is main's tip `61fa7fe8`, and that commit's own `Documentation` run, [35187942641](https://github.com/JuliaGNI/GeometricMachineLearning.jl/actions/runs/35187942641), failed with **66** `##[error]` lines. Both causes are the ones this PR body names, and neither names this page: a `grep -icE "sympnet_tutorial|hamiltonian_neural_network"` over that log returns 0. Six consecutive runs on main are red.

So the test when this branch's run finishes is: any error naming `tutorials/hamiltonian_neural_network.md`, or a count above 66, belongs to this branch.

`Doctests` passed in 1m11s. The CI test matrix is not a gate here, because the diff touches nothing under `src/` or `test/`.

### Also verified

- The training path the tutorial uses is already covered: `test/losses/training_method_losses.jl:122-132` runs the same optimiser call with a `SymplecticEulerLoss`. The tutorial is not the first exercise of it.
- Both losses divide by a norm, at `src/loss/symplectic_euler_loss.jl:90` and `src/loss/hnn_loss.jl:41`. "Why both start near 1" is right.
- `import Random` resolves from `@stdlib` although `Random` is absent from `docs/Project.toml`. The sympnet page already relies on this and builds.
- The changelog entry sits in `### Documentation` of `## [Unreleased] — 0.8.0`, which is the right section. The split into two points is justified, because the seed changes the pre-existing figure too.

PR #282, opened 2026-09-17
Close the defects the scripts gate found

michakraus:
Closes the items `Tasks/Close what the GeometricMachineLearning script gate found.md` lists under **A**, **B1**, **B2** and **D2**. Each was surfaced by the CI gate built in #280 and deliberately left unfixed there.

## What changed

**`SKIPPED` is empty, so all 25 entry points run.** It held two.

`reproduction/linear_symplectic_transformer_gpu.jl` is **repaired**. The coverage that would have justified deleting it is not there: it is the only script that builds a `LinearSymplecticTransformer`, and the sibling it was weighed against builds a `SymplecticTransformer` — a different architecture — and no `GSympNet`. It now takes `backend = smoke_size(CUDABackend(), CPU())` like its four siblings, with `T = Float32` named once, the `cu` on the input made conditional, and smoke sizes on `n_init_con` and `batch_size`: 1000 `ImplicitMidpoint` trajectories was the cost here, not the training.

`reproduction/sympnets/sympnet_pendulum_cuda.jl` is **deleted**. Repairing the `lines!` call it stopped at would have left a file that still cannot run on a GPU. Past that point it is broken in four further independent places: `Lux.setup(CUDA.device(), rng, model)`, where `setup` takes `(rng, model)`; `optimization_step!(opt, model, ps, dp)`, whose second argument is a `GlobalSection` now; `q_learned[0] = q[0]` on a 1-based array, which cannot ever have run; and a `@cuda threads=length(q)` kernel writing `z[i], z[i+1]` into a 2-element `CUDA.zeros(2)`. Its CPU sibling is a full port of the same script and runs.

**Two defects the gate found in passing.** `volume_preserving_feedforward/abc_flow.jl`'s `make_validation_plot(t_validation, nn)` never used `nn` — the body read the global `nn₁`, harmless only because both call sites passed `nn₁`. And `test/data_loader/batch_data_loader_qp_test.jl` contained no `@test` at all, so its testset reported `Total 0`; it reports `Pass 10, Total 10` now.

Four of those ten state the shape bookkeeping that is the file's subject and that nothing else under `test/data_loader/` covers: a `(q, p)` pair of *matrices* is one trajectory, so the second axis is time and `n_params` is 1, where a *bare* matrix of the same shape has its second axis read as the parameter index. Six state that both shapes reach the training loop.

[...]

PR #283, opened 2026-09-17
Make the documentation build green, and vectorise the logo

michakraus:
Closes **C2** and **C1** from `Tasks/Close what the GeometricMachineLearning script gate found.md`, and takes the logo off `gh-pages` while it is there.

The `Documentation` job has been red since **2026-08-26**, for three independent causes. `docs/make.jl` passes no `warnonly`, so it defaults to `Symbol[]` and each is fatal: the build ends `encountered errors [:cross_references, :example_block]`.

## A — 62 `invalid local image` errors, and not one wrong path

All 62 name a `tikz/*.png`. Every one of those files is a build product: `.gitignore` carries a bare `*.png`, and `git ls-files docs/src/tikz` returns 68 tracked files of which **zero** are PNGs. Every name has a tracked `.tex` source and a line in `docs/src/tikz/Makefile`.

The workflow simply stopped building them. [`4281732c`](https://github.com/JuliaGNI/GeometricMachineLearning.jl/commit/4281732c) ("Unify the shared GitHub workflows", 2026-08-31) replaced this repository's `Documenter.yml` with the canonical body and removed the TeX-toolchain install and `make all -C docs/src/tikz` with it. Both are back.

**The make step runs after the documentation environment, not before it, and that ordering is load-bearing.** I placed it after `checkout` first, copying `GeometricOptimizers`, and `make` exited 2:

```
ERROR: LoadError: ArgumentError: Package CairoMakie [13f3f980-...] is required but does not seem to be installed
! LaTeX Error: File `rosenbrock_naked' not found.
I/O Error: Couldn't open file 'grassmann_sampling_light.pdf': No such file or directory
make[1]: *** [png] Error 1
```

`grassmann_sampling_light.tex` and `_dark.tex` call `\immediate\write18{julia --project="../.." rosenbrock_plot.jl}` through xelatex's shell-escape to draw a Makie surface, and `../..` from `docs/src/tikz` is `docs/`. The workflow this repository had before the unification got the order right; the comment at the step now says why, because the obvious tidy-up is to move it up beside the apt install.

**This needs a change outside the repository, and it has already landed.** `GeometricMachineLearning` is now in `DOCS_EXCEPTIONS` in `Knowledge/AI/githooks/install-workflows.sh` and in `DOCS_ADDITIONS` in `verify-workflows.jl`, beside `GeometricOptimizers` and `SimpleSolvers`, which have the same figure pipeline. Without it the next `--apply` deletes these steps again.

The file's **header comment** has been rewritten to say all of this. It inherited the canonical one, which states that the file is a verbatim copy and names two other repositories as the only exceptions — both now false of it. `verify-workflows.jl` ignores comments and checks the behavioural lines instead, which is exactly why a rewritten header is the right place to record the difference; `SimpleSolvers` already does it this way.

## B — an `@example` calling `Optimizer` on a bare `NamedTuple`

`optimizer_comparison.md:30` built `weights = (xy = init_con, )`. `Optimizer` takes a `NeuralNetwork` (`src/optimizers/optimizer.jl:211`) or a `NetworkParameters` (`:217`). Wrapped in the latter. The block ends with `fig` now, so the page shows its surface rather than the `repr` of an `Optimizer`.

## C — two `@ref`s missing their backticks

`[solve!](@ref)` in `symplectic_autoencoder.md` and `[GeometricMachineLearning.ResNetLayer](@ref)` in `transformer.md`. Unbackticked, Documenter looks for a *section* of that title, finds none, and fails `cross_references`. Both names are documented and their backticked forms elsewhere on the same pages resolve.

The last green run was 2026-08-25 and the first red 2026-08-26, so **B and C predate A by five days**.

## `docs/check_references.jl` runs again, and reaches further

The step that invoked it went with the unification, leaving the script tracked and called by nothing. It is the cheap half of this job — `@docs` and code `@ref` resolution need only the loaded package, so it fails in seconds where the build takes the better part of an hour.

It reported `0 unresolved` while the build was dying on two of them, because `ref_targets` matches ``[`x`](@ref)`` and nothing else. It now also reads unbackticked `[text](@ref)`, and flags one only when the text **matches no markdown heading** *and* **does resolve as a documented binding**. That pair of conditions is what keeps it quiet about `[Some Section](@ref)`, which is the legitimate way to link a section by title.

Checked by reverting both fixes rather than by trusting a zero: it exits 1 and names `architectures/transformer.md:15` and `tutorials/symplectic_autoencoder.md:114`.

**It is a fail-fast pass, not a replacement for the build, and the second condition is what bounds it.** A `[text](@ref)` that names neither a heading nor a documented binding — a mistyped or renamed *section* title — is the other way Documenter reports `Cannot resolve @ref`, and this script stays quiet about it. Catching that would mean re-implementing Documenter's anchor map, including the duplicate-heading ambiguity a set of titles cannot represent. `@example` blocks are likewise only checked by the build itself.

## The logo is vector

The `logo` target rendered three PNGs at 500 dpi and copied them into `docs/src/assets/` — 4880×1892 pixels, `1594077`, `1599170` and `1587689` bytes, for a sidebar image about 240 px wide.

| | on disk | over the wire |
|:--|--:|--:|
| `logo.png` (before) | 1 594 077 | 1 594 077 |
| `logo.svg` (after) | 222 480 | **30 986** |

Documenter searches `assets/logo.{svg,png,webp,gif,jpg,jpeg}` in that order, so the SVG is found first, and GitHub Pages serves it gzipped. The two HTML logos go from about **3.19 MB to about 62 kB**, and are sharp at any size instead of at one.

The LaTeX manual takes `logo_without_jl.pdf` — 15 225 bytes — because xelatex includes PDF natively and cannot include SVG. That is not a guess: `assets/logo_tum.pdf` already reaches that build through `\includegraphics` in two places.

`.gitignore` names the three generated files **exactly** rather than sweeping `*.svg` and `*.pdf`, because the tracked `logo_tum.pdf` sits beside them. It also covers `/docs/src/tikz/*.pdf` now — `make` writes one PDF per `.tex`, there are **65** `.tex` sources, and `clean` removes the PDFs at the end of a full run, so an interrupted run left 65 untracked files behind.

The Makefile's `empty` target now removes the three logo outputs it writes into `../assets`. It swept only `../assets/*.png`, which was every logo it wrote before this change and none of them after. The three are named rather than swept, because `rm -f ../assets/*.pdf` would delete the tracked `logo_tum.pdf`.

## C1 — the `ReducedLoss` block stays a plain fence, and the measurement is why

The task asked to check feasibility before converting it to an `@example`. Three facts decide it against:

- `integrator_train_epochs = 65536`, so an `@example` would need a token epoch count;
- the block immediately after already loads the **committed** weights, so a network trained during the build is discarded on the next line;
- the path is already gated by `test/losses/reduced_loss_optimization.jl`, which trains `ReducedLoss` through the `Optimizer` functor on CPU — and whose own header comment names the exact gap C1 describes.

So it would add training to an hour-long job and establish only what a test establishes in seconds. I took the task's stated alternative — a note rather than silence — and made it more useful than "untested": one callout per page, in the pages' own `Main.remark` idiom, covering all six plain fences in the tutorials, saying that they are shown and not run, what the build uses instead, and *where* the path is covered.

## Review round

Rebased onto `f2820dc9`, which added #281 and #282. The only conflict was `CHANGELOG.md`, an add/add under the same `### Documentation` heading; both sides are kept.

- The **volume-preserving-transformer callout was never committed**. The CHANGELOG described one callout per page and only the symplectic-autoencoder page had one. Added.
- The symplectic-autoencoder callout said the committed result follows **every** plain fence. The block after the first defines `mtc`, and the `.h5` load arrives after the next fence. It now says what the build uses rather than what follows each block.
- Two wrong counts for the same thing — `68` in `.gitignore`, `63` in the CHANGELOG. There are **65** `.tex` sources. `68` is the count of *tracked* files in `docs/src/tikz`, which is a different statistic used correctly elsewhere.
- `.gitignore` carried a fourth pattern, `/docs/src/tikz/*.svg`, in a block whose comment says there are three exact ones and no sweep. Nothing writes an SVG there, so the pattern is gone rather than the sentence.
- **`check_references.jl` read every `.md` under `docs/src`.** Two of the 52 are in no `pages =` entry, so Documenter never sees them: their headings could excuse a section `@ref` the build rejects, and their own references were checked against a manual that does not contain them. The list is now read off `make.jl`'s syntax tree, since `_html_pages` and `_latex_pages` are assembled from nested variables and a copy would drift. 50 named, all 50 present, and an empty extraction raises rather than passing blindly.
- **Fence detection matched ```` ``` ```` only at column zero.** A fence indented inside a list or an admonition was read as ordinary text: a matched indented pair leaked the block's contents into the scan, and a pair with one side indented left the toggle inverted and dropped every remaining line. `unbackticked_ref_targets` tracked fences not at all. Both now share one whitespace-tolerant scan.

**The job timeout stays at the canonical 90 minutes**, against a measured 1h08m40s. Raising it was raised in review, but `timeout-minutes` is a behavioural line, and `verify-workflows.jl` requires a `DOCS_ADDITIONS` copy to add steps and change nothing else. Changing it means changing the shared template.

## Verification

The full build, in a clean copy of this branch, before the rebase:

- `docs/make.jl` → **exit 0**, `invalid local` count **0**, zero errors in the log.
- `build/tikz/` holds the manual's figures — the `png` target has 62 `pdftocairo` rules, the three `logo_with_name_*` sources having become SVG and PDF; `build/index.html` carries `` and its dark sibling; **no PNG logo reaches `build/assets/` at all**.
- Rebuilt again after adding the first callout: exit 0, and it renders.
- `make all -C docs/src/tikz` → exit 0 with the corrected step order, against exit 2 before it.
- CI run [35198735386](https://github.com/JuliaGNI/GeometricMachineLearning.jl/actions/runs/35198735386) → **success in 1h08m40s**, the first green `Documentation` run since 2026-08-25.

After the rebase and the review round:

- `docs/check_references.jl` → `0 unresolved`, **exit 0**; and exit 1 naming both sites when the two `@ref` fixes are reverted.
- Both fence failures reproduced against the old code on probe pages before the fix, then gone after it.
- Page extraction measured: 50 of 52 files named in `make.jl`, all present, the two excluded being `data_loader/TODO.md` and `tutorials/softmax_comparison.md`.
- The canonical workflow body survives as a subsequence of this copy — 40 behavioural lines inside 48 — so `documenter_additions_intact` passes.
- `docs/check_references.jl` passes JuliaFormatter under the repo's `sciml` config; all touched files are NFC.

**Not checked.** The LaTeX pipeline (`Latex.yml`) was not run; it is `disabled_manually` and last succeeded before August. The PDF logo takes the same `\includegraphics` path `logo_tum.pdf` already uses, but the manual was not built. **The full documentation build has not been run since the rebase** — CI is running it now.


michakraus (commented):
## Review — `cd40320c`

Reviewed at `origin/T22-documentation-build` = `cd40320c`, rebased onto `f2820dc9`.

**Disclosure: I am no longer a neutral reviewer of this branch.** I found the points below, and then pushed the fixes for them myself at the author's request (`d87506ee`, `a5afe9c2`, `cd40320c`). That makes me an author, so **I will not merge this PR**, and the fixes I wrote need someone else's eyes.

### What the review found, and what became of it

| # | severity | where | finding | state |
|--:|:--|:--|:--|:--|
| 1 | blocker | `docs/src/tutorials/volume_preserving_transformer_rigid_body.md` | The second `Main.remark` callout the CHANGELOG and body both describe was **not on the branch** — it existed only as an uncommitted worktree edit. The CHANGELOG said "one callout per page now does"; one page had none. | fixed, `d87506ee` |
| 2 | blocker | `CHANGELOG.md` | `mergeable=CONFLICTING`. Add/add against #281's own `### Documentation` bullet. | fixed by rebase; now `MERGEABLE` |
| 3 | quality | `.gitignore:23`, `CHANGELOG.md` | Two different wrong counts for one thing — `68` and `63`. There are **65** `.tex` sources, and this PR's own CI log wrote 65 PDFs. `68` is the count of *tracked* files in `docs/src/tikz`, used correctly two paragraphs earlier. | fixed, `a5afe9c2` |
| 4 | quality | `docs/src/tikz/Makefile`, `empty` target | `empty` swept only `../assets/*.png`. The `logo` target now writes two SVGs and a PDF there, so `make empty` left them behind. The three are named rather than swept — `../assets/*.pdf` would delete the tracked `logo_tum.pdf`. | fixed, `a5afe9c2` |
| 5 | nit | `.gitignore:22` | The comment says three exact patterns and no sweep; there were four, and `/docs/src/tikz/*.svg` is a sweep that covers nothing — the `logo` target writes both SVGs into `../assets`. | fixed, `a5afe9c2` |
| 6 | quality | `.github/workflows/Documenter.yml:1` | The retained canonical header says the file is copied verbatim and "never a copy in a repository", and names **two** repositories as the only exceptions. Both are now false of it. `verify-workflows.jl:271` states the convention: a `DOCS_ADDITIONS` repository rewrites its header on purpose, because comments are the part the comparison ignores. | fixed, `a5afe9c2`, in the form `SimpleSolvers` uses |
| 7 | nit | `docs/check_references.jl` | The prose said the blind spot was "closed". The pass fires only when the text resolves as a documented binding, so a mistyped or renamed *section* title — no heading, no binding — still reaches Documenter as `Cannot resolve @ref`. | scope now stated in the CHANGELOG, the body and the workflow comment |
| 8 | nit | `docs/check_references.jl`, `markdown_files` | It walked all of `docs/src`. Two of the 52 pages are in no `pages =` entry, so Documenter never reads them: their headings could excuse a section `@ref` the build rejects, and their own refs were checked against a manual that does not contain them. | fixed, `cd40320c` |
| 9 | nit | `docs/check_references.jl`, fence tracking | ```` ``` ```` was matched at column zero only. A fence indented in a list or admonition read as prose: a matched indented pair leaked the block in, a half-indented pair left the toggle inverted and dropped the rest of the file. `unbackticked_ref_targets` tracked fences not at all. | fixed, `cd40320c` |
| 10 | nit | `docs/src/tutorials/symplectic_autoencoder.md` | The callout said the committed result follows **every** plain fence. The block after the first defines `mtc`; the `.h5` load arrives after the *next* fence. | fixed, `a5afe9c2` |
| 11 | nit | `.github/workflows/Documenter.yml:42` | `timeout-minutes: 90` against a measured 1h08m40s, now with a 2m24s make step added. About 20 minutes of headroom. | **not fixed** — see below |

### Why 11 stays open

`timeout-minutes` is a behavioural line. `verify-workflows.jl:288` requires a `DOCS_ADDITIONS` copy to add steps and change nothing else, and raising it here fails `documenter_additions_intact`:

```
MISSING: "        timeout-minutes: 90"
```

Raising it means raising the canonical template, which touches every repository. That is a separate change, not this PR's.

### Verified

- **The PR body's claims check out.** `src/optimizers/optimizer.jl:211` is `Optimizer(::OptimizerMethod, ::NeuralNetwork; …)` and `:217` is the `::NetworkParameters` method — the cited line numbers are exact, and both constructor forms were run against the loaded package. `solve!` is in a `@docs` block at `reduced_order_modeling/symplectic_mor.md:189`, `GeometricMachineLearning.ResNetLayer` at `architectures/neural_network_integrators.md:60`. `grassmann_sampling_light.tex:9` and `_dark.tex:9` do carry `\immediate\write18{julia --project="../.."}`, and they are the only two files that use `write18`, so the step order is load-bearing as described.
- **`.gitignore` keeps `logo_tum.pdf` tracked.** `git check-ignore -v` returns the three logo outputs with their rule lines and no rule for it.
- **Dropping `imagemagick` is safe** — no `convert`/`magick` in the Makefile. **Removing `const buildpath`** is safe — one occurrence repo-wide, the definition.
- **No `Project.toml` churn**, so no `[compat] julia` bump to break the Documenter pin and no comment-eating rewrite.
- `docs/check_references.jl` → `0 unresolved`, exit 0 on the rebased tree; exit 1 naming both sites when the fixes are reverted. Passes JuliaFormatter under the repo's `sciml` config. All touched files NFC.
- Canonical workflow body intact as a subsequence: 40 behavioural lines inside 48.
- Page extraction measured: 50 of 52 named, all present, the two excluded being `data_loader/TODO.md` and `tutorials/softmax_comparison.md`.
- Both fence failures reproduced against the old code, then gone.

### Not verified

- **The documentation build since the rebase.** The green run [35198735386](https://github.com/JuliaGNI/GeometricMachineLearning.jl/actions/runs/35198735386) (1h08m40s, the first since 2026-08-25) was against a merge commit whose base predates #281 and #282. CI is running the current head now.
- **`Latex.yml`.** `disabled_manually`, last succeeded before August, so the PDF logo path is exercised by nothing.
- **The SVG byte figures**, which need xelatex and `pdftocairo`. The arithmetic in the body is self-consistent.
- **The preview deployment** — `push_preview` reported `✘`, so there is no rendered page to check the SVG logo against.

### Pre-existing, mentioned not fixed

- `docs/src/tikz/Makefile:38` still renders `logo.pdf` → `logo.png`, and nothing under `docs/` references `tikz/logo`. It was already unreferenced at `f2820dc9`, so this branch did not create it; the branch only removed its trailing whitespace.
- `docs/Makefile:34` `install_brenier_two_fluid` calls `Pkg.add(url = …)` although `docs/Project.toml` already carries `BrenierTwoFluid` with a `[sources]` `rev = "main"` pin. The restored Documenter workflow bypasses it; `Latex.yml:33` still calls it.
- `Documentation` is not a required check, so a red docs job does not block a merge here.

**Verdict: the two blockers are cleared and the branch is `MERGEABLE`.** It needs a reviewer who did not write the fixes, and a green `Documentation` run on the current head.

PR #284, opened 2026-09-17
Give the package an Aqua baseline

michakraus:
Closes **B3** from `Tasks/Close what the GeometricMachineLearning script gate found.md`.

Aqua had never run here. It was in neither `Project.toml` nor `test/`, so this package has had **no package-level baseline at all** — no ambiguity check, no stale-dependency check, no compat check, nothing.

## What runs, and what does not

Six of the eight checks `Aqua.test_all` enables by default now run in the suite, reporting `Pass 9, Total 9` between them — `deps_compat` emits four assertions and the other five one each. A piracy-count gate adds one more, so the testset totals `Pass 10, Total 10`.

| check | verdict |
|:--|:--|
| `undefined_exports`, `project_extras`, `stale_deps`, `deps_compat`, `persistent_tasks` | **pass** |
| `unbound_args` | **passes on `min`, `1` and `pre`; fails on nightly** — recorded as *B9* |
| `ambiguities` | **23** — switched off by name, recorded as *B8* |
| `piracies` | **13** — switched off by name, recorded as *B7*, and gated on the count |

**`persistent_tasks` passes.** The known false failure of that check comes from a TestRunner shim's one-entry `JULIA_LOAD_PATH`; a plain `Pkg.test()` does not have one.

**`unbound_args` fails on nightly**, over `Base.iterate(nn::NeuralNetwork{<:NeuralNetworkIntegrator}, ics::BT; n_points)` at `src/architectures/neural_network_integrator.jl:98`. Its `where` clause is `{T, AT <: AbstractVector{T}, BT <: NamedTuple{(:q, :p), Tuple{AT, AT}}}`, and `T` never appears in an argument type. The defect is in the signature and is there on every version; what changes is whether Julia's introspection surfaces it. `nightly` is advisory by construction — `experimental: true` plus job-level `continue-on-error`, deliberately not a required check — so it does not gate this merge. It is *B9* so that the red job has something to match against.

**`deps_compat` needed three lines.** `InteractiveUtils`, `LinearAlgebra` and `Random` were in the root `[deps]` with no `[compat]` bound. Each is `"1"` now, which is what `test/Project.toml` already does for its own stdlibs. With `[compat] julia = "1.11"` the stdlib versions already satisfy `"1"` on the declared floor, so resolution is unchanged.

## Why the two failing checks are disabled by name and not marked `broken = true`

Because a `broken = true` is a check that reports success while the defect stands, and this test suite's two other guards exist precisely to remove that pattern.

The 13 piracies are not a count awaiting triage. **All 13 are genuine under Aqua's definition, and 11 have a value witness** — a call whose behaviour differs between a process holding only the owning packages and the same process with this one loaded. The three on `Base` are the severe ones, because they change *every* Julia process that loads this package:

| method | `src` | without GML | with GML |
|:--|:--|:--|:--|
| `+(::Float64, ::Tuple{Float64})` | `utils.jl:61` | `MethodError` | `3.0` |
| `+(::Vector{Float64}, ::Tuple{Float64})` | `utils.jl:67` | `MethodError` | `3.0` — a **scalar**, discarding every element but the first |
| `isapprox(::@NamedTuple{q, p}, …)` | `utils.jl:163` | `MethodError` | `true` |

The first two already carry a `# Type pyracy!!` comment in the source.

`dim(nn::NeuralNetwork)` at `src/backends/lux.jl:56` turns a clean `MethodError` into a logged error and a returned `nothing`, for every `AbstractNeuralNetworks` user rather than only for users of this package. The subject is the *network*: `dim` on an architecture with no method of its own already logs and returns `nothing` upstream.

**Two of the 13 have no value witness, and the entry says so plainly.** `add!(::SymmetricMatrix, …)` and `add!(C::AbstractVecOrMat, A, B)` both return exactly what the upstream generic returns. Their witness is an allocation regression — the method writes `C .= A + B`, materialising the sum, where upstream writes `x .= a .+ b`. Measured in a fresh process at `--check-bounds=auto`, minimum of 50 calls: **64 bytes for a 1-element vector, 144 at 10, 8256 at 1000, 112 for a 2×2 matrix — against 0 upstream at every size.** Plus a guard weakened from `axes` equality to `size` equality.

Fixing either set is a change to `src/` that this PR does not own. Six checks that fail on a real regression are worth more than eight that are all switched off.

## A switched-off check detects nothing, so `piracies` gets a gate

`test/aqua.jl` asserts `length(Aqua.Piracy.hunt(GeometricMachineLearning)) == 13`. It fails when a piracy is added, and it fails when one is removed without the *B7* entry going with it. Without it, the count and all eleven `src` line numbers in that file's comment would drift unobserved, because a disabled check evaluates nothing.

`ambiguities` gets no such gate. Seventeen of its 23 are against left-multiply methods in ArrayLayouts, FillArrays, Symbolics and GeometricOptimizers, so the count moves with those packages' versions rather than with anything in this tree, and an exact assertion would turn the suite red on an unrelated upgrade.

## Two wiring facts

**`Aqua.test_all` needs an enclosing testset, and `runtests.jl`'s `@safetestset` is it.** `test_all` wraps each check in a `@testset` of its own and adds no parent, and a testset with no parent finalises as soon as it closes — so called bare, the first failing check throws a `TestSetException` and the rest never run. The first measurement of this package reported `ambiguities` and stopped. `test/aqua.jl` keeps a `@testset` as well: it is the parent when that file runs alone, and it is what lets the piracy gate sit beside `test_all` as a second assertion rather than as a top-level `@test` that ends the file when it fails.

**The Aqua block runs after the subject drivers.** A top-level testset throws when it closes on a failure, which ends `runtests.jl`, so whatever runs first can hide everything behind it. `reachability.jl` and `exports.jl` earn the leading position because they say whether the suite is well-formed at all. A package-level finding does not, and placed first it costs every subject result in the run.

**`test/reachability.jl` has an empty `ALLOWED_ORPHANS`**, so `test/aqua.jl` had to be wired into `runtests.jl` or the suite would fail on it.

## Verification

- Full `Pkg.test()` on this branch at the first commit: **exit 0**, 0 failures. `test/aqua.jl` standalone after the review round: `Aqua | Pass 10 Total 10`, exit 0. `test/reachability.jl` standalone: `Pass 2 Total 2`, exit 0.
- 13 piracies and 23 ambiguities measured on this branch, not carried over.
- Every `file:line` in the *B7* table re-derived from `Aqua.Piracy.hunt` and then read in the tree — all 13 match, and each line defines the method the entry names.
- The 23 split re-derived: 17 `PoissonTensor * v` at `poisson_tensor.jl:71,74,77` against ArrayLayouts, FillArrays, Symbolics and GeometricOptimizers; 1 `Dense` vs `Affine`; 1 `_GMLGradient` vs `SimpleSolvers.Gradient`; 4 loss functors vs `NetworkLoss`. 17+1+1+4 = 23.
- The three `[compat]` bounds through Aqua's own `find_missing_deps_compat`: clean with them, exactly those three missing without, and no other `[deps]` entry lacking a bound (18 deps, 20 compat entries).

## Pre-PR verification, and what it caught

**The branch had zero commits when the pass started.** All five files were written and the suite had been run against the working tree, but nothing was committed — a PR opened then would have been empty.

**Three claims were wrong and are corrected:**

- "0 bytes becoming 112" appeared twice as if it were the figure. It is one point on a curve; the spread above replaces it.
- "an architecture that does not implement `dim` raises a clean `MethodError`" blurred the subject: `dim` on an *architecture* already logs and returns `nothing` upstream (`AbstractNeuralNetworks/src/architecture.jl:8`). The `MethodError` belongs to `dim(nn::NeuralNetwork)`.
- "in 19.8 s together" was a machine-specific wall-clock figure; it is replaced by the assertion count, which a reader can reproduce.
- `test/runtests.jl:11` said "The two guards stay here" above a block this branch grew to three.

**Not checked.** Type piracy, instability and allocations are **not attributable to this diff**: it changes no `src/` line. The 13 piracies sit at unchanged lines and are recorded as *B7*.

## Review round

Six findings from the review, all addressed in `9477b9be`:

1. **The PR turned `Julia nightly` from green to red and said so nowhere.** `unbound_args` fails there; the same job is `success` on the base commit `61fa7fe8`, and `test/aqua.jl` does not exist on `main`. Now *B9*, and every assertion count says which Julia versions it belongs to.
2. **The Aqua block ran ahead of the subject drivers**, so its failure blacked out the suite — visible in the nightly log, where nothing from `arrays/`, `layers/` or any other subject ran. It now runs last.
3. **The comment in `test/aqua.jl` called `dim` a `Base` method.** It belongs to `AbstractNeuralNetworks`. `isapprox` at `src/utils.jl:163` is the third `Base` method, and `dim` is named separately.
4. **The `dim` correction announced above had not reached `test/aqua.jl`**, which still read "for every architecture that does not implement `dim`". Fixed.
5. **The comment called the inner `@testset` what "keeps all six checks reachable" standalone**, which was inert with both failing checks off. The piracy gate makes it load-bearing again, and the comment now says why.
6. **Nothing detected drift in the hard-coded `13`, `23` and eleven `src` line numbers.** The piracy gate is that detector; the ambiguity count deliberately gets none.

One candidate finding was checked and rejected: the comment at `src/optimizers/optimizer.jl:24` names `GeometricOptimizers.Gradient` where the ambiguity detector reports `SimpleSolvers`. `GeometricOptimizers.Gradient` resolves to `SimpleSolvers.Gradient` — it is a re-export, and the comment is correct.


michakraus (commented):
## Review

Reviewed at `ee67c5d7` in a detached worktree at origin's tip. The change is sound and the
verification behind it is unusually thorough — every count, every `file:line` and the allocation
curve re-derive exactly. What follows is one substantive item and three factual errors in the new
comment.

### Findings

| # | severity | where | finding |
|--:|:--|:--|:--|
| 1 | medium | `test/aqua.jl:38` | The PR turns `Julia nightly` from green to red, and says so nowhere |
| 2 | medium | `test/runtests.jl:21` | One Aqua failure aborts the whole suite before any subject test runs |
| 3 | medium | `test/aqua.jl:20` | The comment calls `dim` a `Base` method. It is not |
| 4 | low | `test/aqua.jl:26` | A correction the PR body announces did not reach the file |
| 5 | low | `test/aqua.jl:12` | The comment claims the inner `@testset` is load-bearing standalone. It is inert |
| 6 | low | `test/aqua.jl:17` | `13`, `23` and eleven `file:line` are hard-coded with nothing to detect drift |

**1. Nightly goes from green to red, and the entry does not say so.** Aqua's `unbound_args` flags
`Base.iterate(nn::NeuralNetwork{<:NeuralNetworkIntegrator}, ics::BT; n_points)` at
`src/architectures/neural_network_integrator.jl:98`, where `T` appears only inside the bounds of
`AT` and `BT` and never in an argument type. Nightly reports `Aqua's package-level checks | 8 1 9`.

The same job is `success` on the base commit `61fa7fe8`, and `test/aqua.jl` does not exist on
`main`, so this check has never run there. It is **not a merge blocker** — `.github/workflows/CI.yml:22`
makes `pre` and `nightly` advisory (`experimental: true` plus job-level `continue-on-error`), and
branch protection lists seven required contexts, none of them nightly.

The defect is in the claims. The body, the CHANGELOG and the comment all assert `Pass 9, Total 9`
without qualification, on the strength of one local `Pkg.test()` that cannot see nightly. The PR's
own argument is that a check which fails on a real regression beats a check that is switched off.
It has landed such a check, and the entry reads as though nothing fails anywhere.

**2. The Aqua block runs ahead of the subject drivers, so its failure blacks out the suite.** The
nightly log shows the cost: `Reachability`, `Exported names` and the Aqua block produced summaries,
then `ERROR: LoadError` at `test/runtests.jl:21`. Nothing from `arrays/`, `layers/` or any other
subject ran. A package-level check whose failure says nothing about subject correctness should not
be able to hide every subject result.

**3. `dim` is not a `Base` method.** `test/aqua.jl:20` introduces its table as "The severe three are
on `Base`", and the third row is `dim(nn)`. `dim` belongs to `AbstractNeuralNetworks`. The CHANGELOG
table at the same point is correct — `+`, `+` and `isapprox`, all genuinely `Base` — and handles
`dim` separately. The file swapped `isapprox` out for `dim` and kept the claim.

**4. A correction the body announces did not reach the file.** `test/aqua.jl:26` still reads "for
every architecture that does not implement `dim`". The body lists that phrasing as one it corrected,
because the subject is `dim(nn::NeuralNetwork)`: `dim` on an architecture already logs and returns
`nothing` upstream at `AbstractNeuralNetworks/src/architecture.jl:8`. The correction landed in the
CHANGELOG and not here.

**5. The inner `@testset` is not load-bearing standalone.** `test/aqua.jl:12` says it "keeps all six
checks reachable when this file is run on its own". With both failing checks off, a bare
`Aqua.test_all` never throws, so nothing is kept reachable. The body's own pre-PR section records
exactly this measurement — bare and all-passing never throws — and the file did not follow.

**6. Nothing detects drift in the hard-coded detail.** With `ambiguities` and `piracies` off,
neither set is ever evaluated, so a `src/` refactor silently invalidates `13`, `23` and every line
number in the comment. A count assertion is a regression gate rather than the `broken = true` the
entry rightly rejects.

### Verified and correct

Measured on this branch, Julia 1.13.0, Aqua 0.8.16:

- `Aqua.test_all(GML; ambiguities = false, piracies = false)` → `Pass 9, Total 9`, exit 0.
  `deps_compat` really emits four assertions, and 4 + 5×1 = 9.
- `Aqua.Piracy.hunt` → 13. `Test.detect_ambiguities(GML; recursive = true)` → 23, which is what
  `Aqua.test_ambiguities_impl` calls.
- All twelve `file:line` anchors resolve and define the method named.
  `AbstractNeuralNetworks/src/architecture.jl:8` is the only upstream `dim`, which is what makes
  `dim(::NeuralNetwork)` a `MethodError` without this package.
- The B8 split: 17 + 1 + 1 + 4 = 23, each group attributed correctly. "Against `Affine`" is right
  even though the method prints as `Dense`, because `Affine` is a `Dense` alias.
- The three `[compat]` bounds are exactly what `deps_compat` needed — deleting them reproduces
  exactly those three as missing and no other. Resolution is unchanged: the resolved manifest
  entries for all three carry no `version` field at all, so `"1"` cannot bind.
- `ALLOWED_ORPHANS` at `test/reachability.jl:19` is an empty `Dict`, so `test/aqua.jl` did have to
  be wired in.
- "Nine total, eight by default, six here" is right and the two sentences agree. `Base.kwarg_decl`
  on `Aqua.test_all` gives nine names, with `undocumented_names = false`.
- The `add!` allocation curve reproduces byte for byte in cold processes: 64, 144, 8256, 112 against
  0 upstream. The `axes` → `size` weakening is real.
- CHANGELOG placement and conventions are right, and no version bump is correct for a feature PR.
  Formatter and NFC clean on all five files.

One candidate finding was checked and rejected: the comment at `src/optimizers/optimizer.jl:24`
names `GeometricOptimizers.Gradient` where the ambiguity detector reports `SimpleSolvers`.
`GeometricOptimizers.Gradient` resolves to `SimpleSolvers.Gradient` — it is a re-export. The comment
is correct.

### CI

Six of seven required checks pass. `Julia min - ubuntu-latest` is still running. `Documentation`
fails on `[:cross_references, :example_block]`, which is **pre-existing** — red on `main` across
four recent commits — and is not a required check.

PR #285, opened 2026-09-17
Retire the legacy HNN tree, and keep what it had that the package did not

michakraus:
Closes **E1** and **E2** from `Tasks/Close what the GeometricMachineLearning script gate found.md`, and with them **C9** in *Open Issues*.

The direction was: check what the legacy HNN code has that the package does not, integrate the missing part, then delete the rest — and in particular all the Flux and Lux code, with the ModelingToolkit route replaced by `SymbolicNeuralNetworks`.

## The audit came first

Sixteen capabilities across the 13 `.jl` files: **seven covered, eight obsolete, one missing.**

**Covered** — the scalar-network-plus-`𝕁∇H` architecture, by `src/architectures/hamiltonian_neural_network.jl:125-135` with the identical topology; the loss, by `src/loss/hnn_loss.jl`, on a relative norm where the legacy summed squares; the loss gradient by *both* routes, `Zygote` and symbolic; mini-batch selection, by `Batch`, which partitions an epoch where the legacy sampled with replacement; the contour plot; and the pendulum data and its `H`/`∇H`/`dH`, by `scripts/utilities/pendulum.jl`.

**Obsolete** — obsolete for the framework rather than for the mathematics: the hand-rolled SGD loops, the `init_adam`/`apply!` wiring, the Flux front end, the manual `Lux.setup` and its `NamedTuple`→`Tuple` conversion, the `@generated Lux.applychain` workaround for nested `Zygote` differentiation (which the package sidesteps entirely by taking the second derivative symbolically), the Metal port, the ModelingToolkit generation step, and the hard-wired three-layer MLP with its `expand` flattening.

**`legacy/mtk/` maps four for four onto `SymbolicNeuralNetworks`**: `est.jl` is the network itself, `field.jl` is `hamiltonian_vector_field`, `loss.jl` is `HNNLoss`, `step.jl` is `SymbolicPullback`. Two real differences, neither arguing for keeping it: the MTK route was frozen at `n_in = 2`, `ld = 5`, `tanh` and three layers where `StandardHamiltonianArchitecture` is generic in all four; and it wrote generated source to disk — `step.jl` is 896 kB of scalarised expression — where `SymbolicNeuralNetworks` builds in memory. `generate_hnn_mt.jl` could not be re-run in any case: it calls the removed `@derivatives` macro.

## The one missing capability is added, not dropped

`hnn_mt_sim.jl` trained a network and then **stepped** with it, comparing the trajectory and the energy drift against the true Hamiltonian. `hnn_pendulum.jl` plots what the network learned on a grid; the two questions come apart, because a network can match `H` closely on the grid and still drift over a long integration.

Nothing else asked it. `plot_network_sim` was the plotting half and had **no caller outside `legacy/`**, and no test touches energy conservation. Deleting the directory without this would have orphaned 57 lines and removed the only executable record of the comparison.

`scripts/reproduction/hnn_pendulum_simulation.jl` is that script. It is discovered by `runscripts.jl`'s directory walk with no registration, so CI gates it from the moment it lands.

**`plot_network_sim` takes the two solutions now instead of integrating them.** Its old body called `integrate(ODE(v, x₀), TableauGLRK(2), Δt, nt)` with `v(t, x, v)` — an interface that predates `ODEProblem(v, timespan, timestep, q₀)` and `v(v, t, q, params)`, and `TableauGLRK` is absent from the API of all seven installed `GeometricIntegrators` versions — `grep -rn TableauGLRK ~/.julia/packages/GeometricIntegrators/*/src/` returns nothing, though the name survives in `prototyping/` in three of them. Since the function had no working caller, nothing had executed that code in a long time. Moving the integration into the script puts the choice of method and step size where it belongs and leaves the utility needing no solver.

One trap it hides, and the first run failed on it: **a comprehension over a `DataSeries` keeps that series' 0-based axes, while `collect` on the same series returns a 1-based vector.** Broadcasting the two is a `DimensionMismatch`, so the drift is built through `parent`.

## Then the deletion

17 files, of which 14 are Julia and 866 lines. **This closes C9**: all 28 of the string-literal `include` sites it counted were in these two directories — 26 under `hnn`, 2 under `mtk` — so the `data.jl` question it waited on is moot rather than answered. Zero remain.

Two findings worth more than the code was:

- **`hnn_lux_metal.jl:57-58` commented the Poisson tensor out and returned the plain gradient.** That file was not an HNN.
- **The legacy data was inconsistent with the legacy loss.** `data.jl` set `target = ∇H.(data)`, the plain gradient, while every other loss compared against `𝕁∇H`. `scripts/utilities/pendulum.jl:38` uses `dH.(points)` — the current file fixed a bug, not merely reshaped the data.

## What is *not* claimed

**Flux and ModelingToolkit are gone** — neither name appears anywhere outside `CHANGELOG.md`. **Lux is not.** `src/backends/lux.jl:17,26,37` calls `Lux.setup` and `Lux.apply` and is `include`d at `src/GeometricMachineLearning.jl:256`; `legacy/layers/linear_symplectic.jl` subtypes `Lux.AbstractExplicitLayer`; `scripts/Project.toml` declares it. The entry says so rather than overclaiming.

That surfaced something for a separate decision, deliberately not taken here: **`src/backends/lux.jl` is included into the module but `Lux` is in no `[deps]`, so every method in it throws `UndefVarError: Lux` on first call.** Unreachable dead code inside `src/`. Deleting it is a decision, not a tidy-up.

## Verification

- `scripts/reproduction/hnn_pendulum_simulation.jl` at `GML_SMOKE=1`: **exit 0**, writes its figure. Re-run after the verification edits, still exit 0.
- Orphan sweep over the whole repository: zero references to `legacy/hnn`, `legacy/mtk` or any of the 17 deleted basenames outside `CHANGELOG.md`. Zero `include(` in the 12 remaining `legacy/` files.
- `.gitignore` needs no change: no pattern or comment in it mentions `legacy/`, and the `.pdf` comment is scoped to `scripts/`.
- `plot_network_sim`'s only caller is the new script; `plot_hnn` is untouched and both its callers pass the unchanged form.
- Every changelog figure re-derived by command, including 866 lines (`git diff --numstat`; `wc -l` gives 862 because four files lack a trailing newline).
- JuliaFormatter and `fatou lint` clean on both changed files.

## Pre-PR verification

**Fixed after the pass:** "the package's last Flux, **Lux** and ModelingToolkit code" was false for Lux, as above. The audit split was written "six covered, nine obsolete" against a table that is seven, eight and one — two covered capabilities had been merged into one phrase. "every loss compared against `𝕁∇H`" contradicted the `hnn_lux_metal` sentence two earlier. `∇H̃` named a *symplectic* gradient with the plain-gradient prefix, where `scripts/utilities/pendulum.jl` — included into the same scope — fixes `∇H` as the plain gradient and `dH` as `𝕁∇H`; renamed `dH̃`. The `plot_network_sim` docstring carried release history, which belongs in `CHANGELOG.md` alone.

**The merge-order note is now resolved.** `scripts/runscripts.jl:41` said "23 reproduction scripts" and this branch adds a 24th. [#282](https://github.com/JuliaGNI/GeometricMachineLearning.jl/pull/282) has since merged and set the same line to 22 by deleting one. The branch is rebased onto `main` and the line reads **23**, which `git ls-tree -r --name-only HEAD -- scripts/reproduction | grep -c '\.jl$'` confirms.

## Review follow-up

Two findings from the review are fixed in `Answer the review`:

- `hnn_pendulum_simulation.jl` built the symbolic Hamiltonian vector field **twice** — once inside the inline `HNNLoss(hnn)`, once through an explicit `hamiltonian_vector_field(hnn)`. Each call runs `SymbolicNeuralNetworks.Jacobian` and `build_nn_function`. The loss is named now and the integration uses `loss.hvf`, which matches `hnn_pendulum.jl:35`.
- `scripts/utilities/plots.jl:105` said the top row holds "the two contour panels". That describes `plot_hnn`. `plot_network_sim`'s top row is one contour panel and the trajectory panel.

**Not checked.** Type piracy, instability and allocations are not attributable — nothing under `src/` changed, and the diff's only executable code is in `scripts/`, outside every module. No test references either changed file, and no `docs/` page does.


michakraus (commented):
## Review — approve

Reviewed at `3e6f2aee`. The audit behind the deletion holds up on every number I could re-derive by command, and the one capability the PR keeps is genuinely exercised by CI rather than merely committed. Four findings, two of them fixed on the branch; see *Follow-up* below.

GitHub's review states are unavailable to me on this repository, because every PR here is authored by the authenticated account. This comment carries the verdict instead.

### CI

| check | verdict |
|:--|:--|
| `Scripts - ubuntu-latest` | **pass** — and the log shows `ok 136.9s reproduction/hnn_pendulum_simulation.jl` (run 35198838054). CI executed the new script end to end. This is the decisive check. |
| 8 × `Julia …`, `Doctests`, `codecov/patch`, `codecov/project` | pass |
| `Documentation` | **fail, not attributable** |

The `Documentation` failure is `MethodError: no method matching Optimizer(::GradientMethod, ::@NamedTuple{xy::Matrix{Float64}}; step_size::Float64)` at `docs/src/tutorials/optimizer_comparison.md:34`, which terminates `makedocs`. The same workflow is `failure` on `main` for the last four pushes, including the merge base `61fa7fe8`. Nothing in this diff touches `src/optimizers/` or `docs/`.

### Findings

| # | severity | location | finding |
|--:|:--|:--|:--|
| 1 | quality | `scripts/reproduction/hnn_pendulum_simulation.jl:35,41` | The symbolic Hamiltonian vector field is built **twice**. `:35` passes `HNNLoss(hnn)` inline, and `src/loss/hnn_loss.jl:26` shows that constructor is `HNNLoss(hamiltonian_vector_field(arch))`. `:41` then calls `hamiltonian_vector_field(hnn)` again. Each call runs `SymbolicNeuralNetworks.Jacobian` and `build_nn_function` (`src/architectures/hamiltonian_neural_network.jl:119-123`). Naming the loss and reusing `loss.hvf` builds it once, and matches `scripts/reproduction/hnn_pendulum.jl:35`, which names the loss for the same reason. **Fixed.** |
| 2 | quality | `scripts/utilities/plots.jl:105` | Stale comment: "the two contour panels occupy the top 70% of the figure". That is `plot_hnn`, whose top row is two contour panels (`:33`, `:39`). `plot_network_sim`'s top row is one contour panel (`:78`) and the trajectory panel (`:84`). The comment is pre-existing, but it now contradicts the docstring this PR adds at `:60-61`. **Fixed.** |
| 3 | nit | `scripts/reproduction/hnn_pendulum_simulation.jl:37-39` | "not a second derivative taken again here" is true of the quantity and false of the work, given finding 1. **Fixed** with finding 1. |
| 4 | nit | PR description | "`TableauGLRK` is absent from all seven installed `GeometricIntegrators` versions" overstated. It is absent from the API in all seven, but the name survives in `prototyping/` in three. **Fixed** in the description. The `CHANGELOG.md` wording was already the safe one. |

### The merge conflict, and why 23

`scripts/runscripts.jl:41` was the only conflicting line; `CHANGELOG.md` auto-merged.

| tree | reproduction `.jl` | line 41 |
|:--|--:|:--|
| merge base `61fa7fe8` | 23 | 23 |
| PR head `3e6f2aee` | 24 | 24 |
| `main` after #282 merged | 22 | 22 |
| rebased head | **23** | **23** |

Counts from `git ls-tree -r --name-only <rev> -- scripts/reproduction | grep -c '\.jl$'`. #282 is merged, so this was a rebase due now rather than an ordering question. `scripts/README.md` carries no script count, so nothing else needed the number.

### The description's claims, re-derived

Every checkable claim verified except finding 4.

| claim | verdict |
|:--|:--|
| Nothing outside `CHANGELOG.md` references `legacy/hnn`, `legacy/mtk`, or any of the 17 deleted basenames | verified — `grep -rn` over the worktree; 5 hits, all `CHANGELOG.md`. The generic basenames (`Project.toml`, `loss.jl`, `step.jl`, `est.jl`, `field.jl`, `README.md`) match live files only. |
| `plot_network_sim` has exactly one caller, and the signature matches | verified — definition `plots.jl:67`, call `hnn_pendulum_simulation.jl:53`; argument lists identical |
| It had no caller outside `legacy/` before | verified — `git grep -n plot_network_sim 61fa7fe8` → only `legacy/hnn/hnn_mt_sim.jl:75` |
| `plot_hnn` untouched, both callers unchanged | verified — the diff hunk opens at `@@ -54,26 +54,21 @@`, below `plot_hnn`'s `end`; neither caller is in the diff |
| 17 files, 14 Julia, 866 lines; `wc -l` gives 862 for four missing trailing newlines | verified — `git diff --numstat`, and the summed `wc -l` over the 14 blobs |
| 28 string-literal `include` sites, 26 under `hnn` and 2 under `mtk`, zero remaining | verified — `git grep -c 'include("' 61fa7fe8` gives 29, of which one is prose in `legacy/hnn/Project.toml:2`; the 28 `.jl` sites split 26/2. `grep -rn "include(" legacy/` on HEAD is empty over the 12 remaining files. |
| `Flux` and `ModelingToolkit` appear nowhere outside `CHANGELOG.md` | verified — `grep -rnw`; the other `Flux` substring hits are the `FluxML/Zygote.jl` URLs |
| Lux *is* still present, and in no `[deps]` | verified — `src/backends/lux.jl:17,26,37`, included at `src/GeometricMachineLearning.jl:256`, `Lux` absent from `Project.toml`. The dead-code observation holds. |
| `hnn_lux_metal.jl:57-58` commented the Poisson tensor out | verified against the merge-base blob |
| `step.jl` is 896 kB; deleting would have orphaned 57 lines | verified — `git cat-file -s` gives 895831; the old `plot_network_sim` spans `:57`→EOF |
| The script is discovered by the directory walk with no registration | verified — it is in no list in `runscripts.jl`, and CI ran it |
| `.gitignore` needs no change | verified — `:31` is `scripts/**/*.pdf`; no `legacy` anywhere in the file |
| No test and no `docs/` page references either changed file | verified — `grep -rn` over `test/`, `docs/`, `.github/` is empty |
| JuliaFormatter and `fatou lint` clean on the changed files | verified, re-run after the fixes |

One caveat on a claim that is true but uncheckable from the repository: `data.jl` set `target = ∇H.(data)`, the plain gradient. That file left the tree long ago and is not one of the 17 deleted files, so a reader has to go to `cc1d7e8a^` to see it.

### Correctness of the new code

- **The integration uses the learned Hamiltonian, and the same object the loss trains against.** `hamiltonian_vector_field` returns `𝕁 * ∇H̃` (`src/architectures/hamiltonian_neural_network.jl:89-104`), and `HNNLoss` compares that against the data (`src/loss/hnn_loss.jl:26,41`).
- **The two fields share the sign convention.** `scripts/utilities/pendulum.jl:11` is `dH(x) = [0 1; -1 0] * ∇H(x)`, which for `n = 1` is the package's `𝕁`. Without this the trajectories would counter-rotate and the comparison would say nothing.
- **The `DataSeries` claims are right.** `GeometricSolutions/src/dataseries.jl:43` defines `getindex(ds, ::Colon, j)` as an `OffsetArray` over `eachindex(ds)`, so `sol.q[:, k]` is the component series and it does carry 0-based axes. The old `sol.q[1, :]` hit `getindex(ds, ::Int, i)` instead — the state at one time index. The old code plotted two points; the new code plots a trajectory.
- **The drift is computed correctly.** Each trajectory is taken against its **own** initial value. The learned Hamiltonian is determined only up to an additive constant, so the drift is what is comparable, not the value. `parent` is a no-op if the comprehension ever returns a plain `Vector`, so it is safe either way.
- **The utility needs no solver now.** `plots.jl:1-2` imports only `CairoMakie` and `LaTeXStrings`.

### CHANGELOG

Right convention and right section. Every line reference in it resolves exactly: `src/architectures/hamiltonian_neural_network.jl:125-135` is `Chain(arch::HamiltonianArchitecture)` through its `end`, `scripts/utilities/pendulum.jl:4-11` is `H` through `dH`, `:30-41` is `get_data_set`, and the three `src/backends/lux.jl` lines are the three `Lux.setup`/`Lux.apply` calls. It **under-claims where the title over-claims**, and says so: Lux is not gone, and the entry states why. Closing C9 by vacating the number matches the register's own convention — C3-C5, C8, C10 and C11 are already vacant.

One observation, not a finding: `legacy/` is not part of the shipped package, so this removal is not breaking for a user. It sits under `### Removed (breaking)` alongside the `scripts/test/` removal, which is equally non-breaking. The 0.8.0 entry as a whole is breaking and the heading is pre-existing, so I would leave it.

### Not checked

- The figure's scientific content. A smoke run trains 2 epochs, so no publication-quality drift curve was produced here.
- Type piracy, inference and allocations. Nothing under `src/` changed, and the diff's only executable code is under `scripts/`, outside every module.
- The rest of the `[Unreleased] — 0.8.0` entry, which is not this branch's work.

### Follow-up

The branch is rebased onto `main` at `2ac4e7e7`, the conflict is resolved to 23, and findings 1-3 are fixed in `Answer the review`. JuliaFormatter, `fatou lint` and the NFC check are clean on all three touched files; the one `fatou lint` warning in `scripts/runscripts.jl:159` is the driver's own `exit` and is present unchanged on `main`.

PR #294, opened 2026-09-18
Add a positional encoding layer, and end legacy/

michakraus:
`PositionalEncoding` adds the sinusoidal encoding of [vaswani2017attention] to its input;
`positional_encoding(T, dim, seq_length)` builds the matrix. Attention is permutation-equivariant —
measured at `2.2e-16` for `MultiHeadAttention` and `4.4e-16` for a whole `Transformer` — so on its
own it cannot tell one ordering of a sequence from another. This is the standard remedy.

Its last input, `legacy/embeddings/sin_cos.jl`, is the reason: rather than delete the file or leave
it sitting unused, it is promoted into `src/`. **`legacy/` no longer exists.** It held thirty files
at `v0.7.0`.

## Three design choices

- **It is off by default.** `Transformer` gains `positional_encoding::Bool = false`, which puts the
  layer at the front of the chain. The transformers here are usually applied to phase-space
  trajectories, where the ordering is carried by the data.
- **It stores no sequence length**, reading it from the input's second axis at every call, because a
  network here is applied to trajectories of whatever length the data has. The cost is one
  `dim × seq_length` allocation per call — measured at 8272 B for `32 × 64`, fully accounted.
- **It is written from the paper, not carried over.** The legacy version paired its rows off by one:
  its first row was an unpaired cosine, so rows 2 and 3 shared a frequency instead of rows 1 and 2.
  Reproduced from the deleted file — sum of squares 0.18 for rows 1,2 against 1.0 for rows 2,3 — and
  the test discriminates on exactly that.

## It does not disturb the geometry

The layer adds a constant, so its Jacobian is the identity, so a structure-preserving architecture
with it in front preserves what it preserved without it. The test asserts that directly, and not
with `==`: the identity `(x + δ + P) − (x + P) = δ` is exact in ℝ and not in floating point, because
each addition rounds and the two round differently. The bound is eight machine epsilons; measured
deviation `4.4e-16` against `1.78e-15`, so the bound is tight rather than slack.

## Known limitation

`positional_encoding` builds a `Matrix`, so the layer adds a **host** array to whatever it is given,
where every other layer here allocates through `KernelAbstractions.allocate` or `similar(x, …)`. A
network carrying this layer is a CPU network, and nothing in the suite would catch that, because the
suite has no GPU test. The docstring carries this as a warning.

## Pre-PR verification

`julia-branch-verifier`, against `origin/main...HEAD`. **It found a functional defect, not only
prose:**

- **The keyword removed an input type the Transformer otherwise accepts.** `Transformer(4,2,1)` takes
  a `(q, p)` `NamedTuple`; with `positional_encoding = true` it raised `MethodError`, because
  `MultiHeadAttention` and `ResNetLayer` both carry a `(z::QPT, ps::NamedTuple)` method and this
  layer did not. The verifier called the fix a design decision; it is not — both siblings stack
  `vcat(z.q, z.p)`, apply at the full dimension and split with `assign_q_and_p`, so this layer now
  does the same. The test asserts both that `(q, p)` is accepted and that the stacked and split
  paths agree exactly.

Four claims of mine were false and are corrected:

- `legacy/` held **thirty** files at `v0.7.0`, not the twelve the entry said — twelve was a mid-cycle
  count, after `legacy/hnn/` and `legacy/mtk/` went.
- The wavelengths reach `10000·2π` only as the dimension grows: at `d = 4` the longest is `100·2π`,
  and the tests use `dim` 2–8.
- "Asserted in the test suite" followed two composition claims no test covers. The suite asserts the
  Jacobian identity; the compositions follow from it.
- "Positions are counted from zero, **as in the paper**" — Vaswani et al. §3.5 gives no base for
  `pos`. The behaviour is right and standard; the attribution was not.

Two further fixes: one assertion restated the functor's body, which the file's own comment warns
against, and two comments referred to `legacy/` — which this branch deletes, and which falsified an
unchanged CHANGELOG sentence in the same section.

**Checked and clean.** The formula recomputed independently from the paper: max deviation `1.7e-15`
over six shapes, including odd `dim`. Type stability: all return types concrete. Piracy and
ambiguity attributable to this diff: zero, against a package baseline of 12 and 23. No reference to
`legacy/` remains anywhere outside `CHANGELOG.md`.

## Checks

`Pkg.test()` — exit status 0, **0 failed**, the 2 pre-existing `@test_broken` unchanged. The new
testset reports **106 assertions**; the reachability guard passes with an empty allowlist and the
exports guard resolves both new names.

`Documenter.doctest` passes — and was **shown to detect a positive**: breaking one expected value in
the new docstring produced `doctest failure in src/layers/positional_encoding.jl:21-33`. It is worth
saying that the docs build does *not* cover this, since `docs/make.jl` sets `doctest = false` and
the doctests run in their own CI job.

The docs build reproduces only the 62 pre-existing `invalid local link/image` errors for gitignored
`tikz/*.png` build products; none names the new page, and there is no `missing_docs` error.


michakraus (commented):
## Review — verdict: request changes

Reviewed at `28e2537a`, the tip of `origin/feature/positional-encoding`, in a detached worktree.
Every figure below was re-measured in that worktree against `test/Project.toml` with
`--check-bounds=auto`; nothing is quoted from the PR body.

The mathematics is right, the formula reproduces the paper, and the legacy off-by-one claim
reproduces exactly as described. One blocker, and it is not a style point: **a network carrying
this layer cannot be trained.**

---

### 1 — blocker: `positional_encoding` is not differentiable, so the layer cannot be trained

`src/layers/positional_encoding.jl:38-49` builds the matrix with `Matrix{T}(undef, …)` and a
`setindex!` loop. Zygote — the AD this package trains with, `test/transformers/transformer_optimizer.jl:1` —
refuses to differentiate that, and it is reached because the layer calls it inside the forward pass
at `src/layers/positional_encoding.jl:107`.

```
$ julia --project=test --check-bounds=auto
using GeometricMachineLearning, KernelAbstractions, Zygote
for kw in (false, true)
    nn = NeuralNetwork(Transformer(4, 2, 1; positional_encoding = kw), CPU(), Float32)
    xx = rand(Float32, 4, 5)
    try
        Zygote.gradient(p -> sum(nn(xx, p)), nn.params)
        println("pe=$kw  gradient ok")
    catch e
        println("pe=$kw  ERROR ", first(sprint(showerror, e), 80))
    end
end
```

```
pe=false  gradient ok
pe=true  ERROR Mutating arrays is not supported -- called setindex!(Matrix{Float32}, ...)
```

The failing frame names the file:

```
[5] (::Zygote.Pullback{Tuple{typeof(positional_encoding), Type{Float32}, Int64, Int64}, Any})(Δ::Matrix{Float32})
    @ src/layers/positional_encoding.jl:107
```

The same failure comes through `FeedForwardLoss`, i.e. through the package's own training path:

```
pe=false  loss gradient ok
pe=true  ERROR Mutating arrays is not supported -- called setindex!(Matrix{Float32}, ...)
```

This is the whole point of the layer, so it has to work before merge. The fix is one line, and it
is already the package's idiom — `ChainRulesCore` is a direct dependency (`Project.toml:8`) and
`src/kernels/kernel_ad_routines/` is full of hand-written rules. Adding

```julia
ChainRulesCore.@non_differentiable positional_encoding(::Any, ::Any, ::Any)
```

makes it pass, and gives the gradient the layer's own Jacobian claim predicts:

```
gradient with @non_differentiable: ok
d/dx all ones (identity Jacobian): true
```

(Verified by declaring the rule in a probe script against the unmodified worktree; no source was
edited.)

**And a test.** No assertion in `test/transformers/positional_encoding.jl` takes a gradient, which
is why a green suite says nothing here. `transformer_gradient_test` in
`test/transformers/transformer_optimizer.jl:10` is exactly the right shape — running it once with
`positional_encoding = true` would have caught this.

---

### 2 — correctness-risk: the CPU-only limitation is a hard failure, and the fix is small

The docstring warning at `src/layers/positional_encoding.jl:63-67` is accurate but understates the
consequence. Broadcasting a device array against a host `Matrix` does not fall back to the host —
it fails to compile:

```julia
using Metal
Metal.mtl(rand(Float32, 4, 3)) .+ rand(Float32, 4, 3)
```

```
GPU compilation of MethodInstance for (::Metal.var"#broadcast_2d#_copyto!##2")(::MtlDeviceMatrix{Float32, 1},
::Base.Broadcast.Broadcasted{…, typeof(+), Tuple{…Extruded{MtlDeviceMatrix{Float32, 1}…},
Base.Broadcast.Extruded{Matrix{Float32}, …
```

So `Transformer(dim, n_heads, L; positional_encoding = true)` on a GPU backend is not a slow
network, it is an `InvalidIRError` at the first forward pass, from a keyword whose docstring reads
as a free improvement. Every other layer here allocates through the backend; this is the one
exception, and it is introduced by this diff rather than inherited.

A docstring warning is the right remedy for a limitation that is expensive to remove. This one is
not expensive: `similar(x, M, size(x, 2))` plus a broadcast over `CartesianIndices`, or
`KernelAbstractions.allocate(get_backend(x), …)`, is a handful of lines and removes the exception
entirely. I would rather see it fixed here than recorded as known. If it stays, the `Transformer`
keyword's own docstring (`src/layers/transformer.jl:23-27`) should carry the warning too — a user
reaches the keyword without ever reading `PositionalEncoding`.

---

### 3 — quality: a `Vector` input is silently promoted to a one-column `Matrix`

`x .+ positional_encoding(eltype(x), M, size(x, 2))` broadcasts a `(M, 1)` matrix against an
`(M,)` vector, so the rank changes. Measured through the architecture:

```
pe=false  Vector  -> ERROR MethodError: no method matching compute_output_of_mha(...)
pe=false  QPT-vec -> ERROR MethodError: no method matching compute_output_of_mha(...)
pe=true   Vector  -> (4, 1)
pe=true   QPT-vec -> ((2, 1), (2, 1))
```

The PR's own stated principle is that the keyword must not *remove* an input type the transformer
otherwise takes. The converse holds as well: it should not silently *add* one by reshaping. A
vector through a transformer is meaningless, so the `MethodError` at `pe=false` is the right
answer, and `pe=true` turning it into a length-1 sequence hides the mistake. Dispatching the
functor on `AbstractMatrix` and `AbstractArray{T, 3}`, as the siblings effectively are, would keep
the `MethodError`.

---

### 4 — quality: the CHANGELOG still carries a claim the docstring corrects

`CHANGELOG.md:1043`:

> Positions are counted from zero, **as in the paper**, so the first column is ``(0, 1, 0, 1, \ldots)``.

The PR body lists this exact attribution among the four false claims it corrects, and the docstring
is corrected — `src/layers/positional_encoding.jl:17-18` now reads "the paper writes only `pos` is
the position and gives no base". The CHANGELOG sentence was not updated with it. The two now
disagree in the same release.

---

### 5 — quality: a comment refers to a comment that is not there

`test/transformers/positional_encoding.jl:50-51`:

> Against a literal rather than against `x .+ positional_encoding(...)`, which would restate the
> functor's body — **the thing the comment at the top of this file warns about.**

There is no comment at the top of that file. Lines 1-8 are `using` statements and `Random.seed!`.
The comment meant is at lines 15-16, inside the first `@testset`. A reader sent to the top of the
file finds nothing.

---

### 6 — nit: a test comment narrates this branch's history

`test/transformers/positional_encoding.jl:92-94`:

> The keyword must not *remove* an input type the transformer otherwise takes. **It did:** without
> this method a `(q, p)` `NamedTuple` raised a `MethodError` with `positional_encoding = true` and
> worked without it.

"It did" is the development history of this branch, which belongs in `CHANGELOG.md` and not in a
test comment — the same rule the rest of this repository is careful about. The first sentence alone
says everything the next reader needs. The source-side version at
`src/layers/positional_encoding.jl:110-113` gets this right: it is a counterfactual about the
method that exists, not a record of one that did not.

---

### 7 — nit: the allocation figure measures the helper, not the layer call

The body says "The cost is one `dim × seq_length` allocation per call — measured at 8272 B for
`32 × 64`". Re-measured, warm, in a fresh process:

| what | measured | dense `Float32` `32 × 64` |
|:--|--:|--:|
| `positional_encoding(Float32, 32, 64)` | 8272 B | 8192 B |
| `PositionalEncoding(32)(x, NamedTuple())` | 16544 B | — |

8272 B reproduces exactly. The layer call is twice that, because the output is allocated as well.
The docstring at line 61 is the careful phrasing — "the encoding matrix is therefore built per
call, which costs one `dim × seq_length` allocation" — and is the marginal cost over a layer that
only allocates its output. Worth saying in the body the way the docstring says it.

---

## Verified and fine

**The formula is the paper's**, recomputed independently from Vaswani et al. §3.5 —
`PE(pos, 2i) = sin(pos / 10000^(2i/d))`, `PE(pos, 2i+1) = cos(…)`, feature index 0-based — rather
than from the implementation:

| `dim` × `seq` | max deviation |
|:--|--:|
| 2 × 5 | 0.0 |
| 3 × 5 | 8.7e-19 |
| 4 × 6 | 6.9e-18 |
| 5 × 7 | 2.8e-17 |
| 7 × 4 | 0.0 |
| 8 × 3 | 3.5e-18 |
| 64 × 128 | 1.4e-14 |

Odd `dim` is included and behaves: at `dim = 5` the last row is an unpaired sine of the truncated
pair, `row5 ≈ sin((pos-1)/10000^(4/5))` → `true`.

**The row pairing, and the legacy off-by-one.** Both reproduced directly, `dim = 6`, as
`max |P[a,:]² + P[b,:]² − 1|`:

| rows | this PR | `legacy/embeddings/sin_cos.jl` |
|:--|--:|--:|
| (1,2) | 1.1e-16 | 0.867 |
| (2,3) | 0.867 | 1.1e-16 |
| (3,4) | 1.1e-16 | 0.131 |
| (4,5) | 0.102 | 1.1e-16 |

The legacy file pairs (2,3) and (4,5) and leaves row 1 an unpaired cosine; this PR pairs (1,2) and
(3,4). The body's description of the defect is exact, and the new test discriminates on it.

**The wavelength claim.** Longest wavelength `= 2π · 10000^(2⌊(d-1)/2⌋/d)`: `d = 4` → `100·2π`,
`d = 8` → `1000·2π`, `d = 64` → `7499·2π`. The docstring and the docs page both state the `d = 4`
value and both say the paper's `10000·2π` is only approached as `d` grows. Correct.

**The `(q, p)` path matches its siblings and agrees with the stacked one.** `ResNetLayer`
(`src/layers/resnet.jl:75-81`) asserts `size(z.q, 1) * 2 == M` and splits at `N ÷ 2`;
`MultiHeadAttention` (`src/layers/multi_head_attention.jl:185-189`) takes the half-dimension from
the input. `src/layers/positional_encoding.jl:114-117` follows `ResNetLayer` exactly. The test at
line 102 asserts the two paths agree with `==`, and it passes.

**Structure preservation.** The Jacobian argument is sound: a translation of phase space is
symplectic and volume-preserving, so `φ ∘ τ` preserves whatever `φ` does. No test composes the
layer with `LinearSymplecticTransformer` or `VolumePreservingTransformer`, and both the docs page
(line 30) and the PR body say so plainly rather than claiming coverage they do not have. That
honesty is right.

**Type stability** — every return type concrete, via `Base.return_types`:

```
positional_encoding(Type{Float32}, Int, Int) -> Matrix{Float32}
layer on Matrix                              -> Matrix{Float32}
layer on QPT                                 -> @NamedTuple{q::Matrix{Float32}, p::Matrix{Float32}}
layer on 3-tensor                            -> Array{Float32, 3}
```

**Piracy and ambiguity, attributable to the diff: zero.** `Aqua.Piracy.hunt` returns 12, which is
the documented gate at `test/aqua.jl:63`; `Test.detect_ambiguities` returns 18, the documented
GML-only baseline at `test/aqua.jl:28`. Neither set names `PositionalEncoding` or
`positional_encoding`. `initialparameters` and `parameterlength` dispatch on the package's own
type, and the signature matches `src/layers/bias_layer.jl:9-10` field for field.

**The new testset runs and passes** — 106 assertions, which is the body's figure:

```
the encoding matrix is the formula of the paper                    | 71
positions are counted from zero, and row pairs share a frequency   | 13
the element type is honoured                                       |  3
the layer adds the encoding and has no parameters                  |  3
the sequence length comes from the input, not the constructor      |  5
the Jacobian is the identity                                       |  1
a (q, p) input is accepted, and agrees with the stacked one         |  5
the Transformer keyword puts the layer at the front, and only then |  5
```

**`legacy/` and its file count.** `git ls-tree -r --name-only v0.7.0 -- legacy/ | wc -l` → `30`;
the same on `origin/pr/294` → `0`. The CHANGELOG's "held thirty files at `v0.7.0` and holds none"
is exact, and it frames the count as the whole release rather than this branch, which is the
honest reading. No reference to `legacy/` survives outside `CHANGELOG.md` (the only other hit,
`test/parameters/hdf5_support.jl:30`, is an unrelated `_legacy_group` helper).

**Registration.** Both names are exported (`src/GeometricMachineLearning.jl:211`), both appear in
the `@docs` block of the new page (`docs/src/layers/positional_encoding.md:37-40`), and the page is
in both the HTML and the LaTeX tables of contents (`docs/make.jl:232` and `:323`, via
`_special_layers`). `vaswani2017attention` resolves in `docs/src/GeometricMachineLearning.bib:183`.
`test/exports.jl` and `test/reachability.jl` both pass. The `Transformer` docstring renders
correctly with the new keyword. All six changed `.jl` files pass JuliaFormatter against the repo's
`sciml` config.

---

## Not checked

- **The full `Pkg.test()` suite.** I ran the new testset, `exports.jl` and `reachability.jl`
  only. CI runs the whole suite on nine platforms and every one of those jobs was still in
  progress when this was written, so I report no matrix result. It does not change the
  verdict: a green suite cannot bear on finding 1, because nothing in the suite takes a
  gradient through this layer — which is the finding.
- **The docs build.** The `Documentation` job was still running when this was written. The
  `Doctests` job is green, which covers both `jldoctest` blocks in the new docstrings.
- **GPU behaviour of the layer itself.** No CUDA or Metal build of the package was available; the
  Metal evidence in finding 2 is the broadcasting mechanism in isolation, not a run of this layer.

---

## Pre-existing, not attributable to this diff — mentioned, not requested

`test/transformers/transformer_optimizer.jl:11-12` assigns `model = Chain(Transformer(…), ResNetLayer(dim))`
and overwrites it on the next line with `model = Transformer(…)`. The first line is dead, and the
`ResNetLayer` import it justifies is then only used there. Since finding 1 asks for a change to
this file anyway, it is cheap to clean up in the same commit.

---

Once finding 1 is fixed with a test that would have caught it, and findings 4 and 5 corrected, this
is ready. Findings 2 and 3 are my recommendation rather than a condition; say so if you disagree
and I will not hold the PR on them.

PR #297, opened 2026-09-18
Audit Part C: allocations, inference and the minibatch copies

michakraus:
Part C of the v0.7.0 whole-package audit: §8 allocations and inference, §9 redundant buffers, §10
the minibatch copies. Parts A ([#292](https://github.com/JuliaGNI/GeometricMachineLearning.jl/pull/292))
and B ([#295](https://github.com/JuliaGNI/GeometricMachineLearning.jl/pull/295)) are merged.

## §8 — the minibatch index set

`batch_over_two_axes` grew a tuple by splatting once per minibatch, so every iteration built a new
tuple type. It inferred as `Tuple{Vararg{Vector{Tuple{Int, Int}}}}` — not concrete — and the
`for batch_indices in batches` loop in `optimize_for_one_epoch!` dispatched dynamically once per
batch. It returns a `Vector{Vector{Tuple{Int, Int}}}` now.

Cold `--check-bounds=auto` processes, baseline and branch measured separately, on
`DataLoader(rand(Float32, 4, n))` with `Batch(8)`:

| columns | minibatches | before | after |
|--:|--:|--:|--:|
| 200 | 25 | 13,904 B, 0.014 ms | 10,464 B, 0.001 ms |
| 2,000 | 250 | 369,824 B, 0.587 ms | 99,440 B, 0.009 ms |
| 8,000 | 1,000 | 4,530,736 B, 41.77 ms | 397,088 B, 0.032 ms |

`test/data_loader/batch_index_set.jl` asserts the inferred return type is concrete and that the
allocation ratio between 250 and 1,000 minibatches stays below six, so neither property can regress
silently.

**This is visible to callers.** The functor returns a vector, not a tuple, so `length.(batch(dl))`
gives `[2, 2, 1]` where it gave `(2, 2, 1)`. The `Batch` and `number_of_batches` docstrings, the
`data_loader` documentation page and the two tests mirroring those doctests are updated. Anything
that iterates the result, takes its `length` or `vcat`s it is unaffected.

## §9 — buffers

**The attention head loop.** Both `compute_output_of_mha` methods concatenated inside the loop, so
the intermediate rows grew with the square of the head count. They build the head outputs first and
`vcat` over all of them once. The discarded `single_head_output` buffer is gone, and the head key is
`Symbol("head_", i)`.

Two formulations that look right do not work here, and both were tried and measured:

- A **preallocated output** written into per head — the shape the audit proposed — raises
  `Mutating arrays is not supported -- called setindex!` under Zygote. These are forward passes, so
  the concatenation has to stay differentiable.
- **`reduce(vcat, …)`** takes its linear path only for `AbstractVecOrMat`, so on the 3-tensor method
  it folds pairwise and stays quadratic. Concatenating 32 head tensors of 4×64×16: 8,646,272 B that
  way against 539,344 B for the variadic `vcat`. At four heads `reduce` measured *better* on the
  real layer, which is why the growth was measured rather than one size.

The variadic call needs its result type asserted, `vcat(head_outputs...)::eltype(head_outputs)`.
Splatting a vector hides the argument count from inference: without it, `compute_output_of_mha`
infers as `Union{Vector{Any}, Vector{Float32}, Matrix{Float32}}` for matrix input and as `Any` for
tensor input, where both were concrete on `main`, and the `Any` propagates out through the layer
functor into the whole forward pass. The assertion costs nothing — allocations are identical at
every size below. **The pre-PR gate caught this; the suite did not, because nothing in `test/`
asserts the inference of a forward pass.**

Allocations per call, checksums and Zygote gradients unchanged at every size:

| dimension, heads, length, data | path | before | after |
|:--|:--|--:|--:|
| 64, 8, 32, 16 | matrix | 176,592 B | 150,272 B |
| 64, 8, 32, 16 | tensor | 2,732,208 B | 2,261,296 B |
| 128, 16, 64, 16 | matrix | 1,221,424 B | 975,168 B |
| 128, 16, 64, 16 | tensor | 19,308,848 B | 15,306,992 B |
| 128, 32, 64, 16 | matrix | 2,311,760 B | 1,802,688 B |
| 128, 32, 64, 16 | tensor | 36,233,776 B | 28,008,560 B |

At 4 heads on an 8×6 matrix the change costs 592 bytes more — 6,944 against 6,352 — which is the
fixed cost of the head vector. The gain grows with the head count: −15 % at 8 heads, −22 % at 32.

**Kernel buffers.** `KernelAbstractions.zeros` → `allocate` in six multiply wrappers, and
`copy`/`zero` → `similar` in `symmetric_mat_mul`, `lo_mat_mul`, `up_mat_mul` and `skew_mat_mul`.
Each kernel was read first: all five write every `C[i, j, l]` over `ndrange = size(C)`, so nothing
reads an uninitialised element.

**The honest size of this is small.** In isolation at 64³ `Float32`, `zeros` costs 31.3 µs against
`allocate`'s 1.08 µs, `copy` 12.4 µs and `zero` 8.5 µs against `similar`'s 0.04 µs. Inside the
multiplies those precede — 3.5 ms to 8 ms at that size — the saving is under half a percent and does
not clear the measurement noise. Do not read a speed-up for the multiplies into this.

Two buffers that look identical are deliberately untouched: `augment_zeros` zeroes a tensor its
kernel writes only a slice of, and the `kernel_ad_routines` buffers are accumulators.

Also removed: an unused `time_indices` comprehension in `convert_input_and_batch_indices_to_array`,
and a duplicated `size(output_diff)` destructure in `augment_zeros`.

## §10 — the minibatch copies

`optimize_for_one_epoch!` copied every minibatch of every epoch under
*"these `copy`s should not be necessary! coming from a Zygote problem!"* — a comment naming no
issue, no version and no reproducer. Zygote has moved from 0.6 to 0.7 since. It does not reproduce:
training runs without the copies for `GSympNet` on matrix and on `(q, p)` data,
`SymplecticAutoencoder`, `StandardTransformerIntegrator`, `LinearSymplecticTransformer`, and a
`DataLoader` carrying a separate output. `_copy` and its three methods are deleted.

`reduced_system.jl` carried the same shape of comment about nested ForwardDiff derivatives. **That
one is not resolved here.** The double evaluation of the decoder is real and visible, but removing
it means restructuring the call around `DiffResults` across a shape change from the `(q, p)`
`NamedTuple` to a flat vector — its own work with its own verification. The comment now states what
is observable and points at new open issue **C14**.

## Verification

| check | result |
|:--|:--|
| `Pkg.test()`, Julia 1.13, `--check-bounds=auto` | 74 testsets, 4302 passing, 2 broken, 0 failed |
| `Documenter.doctest` | passes |
| JuliaFormatter, NFC | clean on every changed file |
| `compute_output_of_mha` on `MetalBackend()` | both paths run; `vcat` returns `MtlArray` of the asserted type, so the assertion does not throw; agrees with CPU to `1.19e-7` and `1.79e-7`, which is `eps(Float32)` |

The 2 broken come from one pre-existing `@test_broken` at
`test/parameters/network_parameters_gradient_structure.jl:41`, identical on `main`.

`fatou lint` reports two pre-existing `redundant-boolean` findings in `src/data_loader/optimize.jl`.
They are not in a hunk this branch changed and are left alone.


michakraus (commented):
## Review — Audit Part C

**Verdict: comment.** The code is correct and I could not break it. Every allocation figure in the
PR body reproduces here *exactly*, measured in cold `--check-bounds=auto` processes against
`origin/main` and `origin/pr/297` separately. Two statements in `CHANGELOG.md` are wrong, though,
and one of them closes an opportunity that is actually open. They are prose only; no source change
is needed.

Reviewed at `d26c612`, from a detached worktree on `origin/pr/297`.

### Findings

| 1 | quality | `CHANGELOG.md` §9, last paragraph | "the `kernel_ad_routines` buffers are accumulators" is not true of any of them. They have exactly the shape §9 just optimised, and the backward pass is where the training time goes. | `src/kernels/kernel_ad_routines/mat_tensor_mul.jl:64`,`:65` allocate `dA = zero(A)` and `dS = KernelAbstractions.zeros(…)`; `:67`,`:68` launch `lower_da!(dA, S, dC, ndrange = size(dA))` and `lower_ds!(dS, A, dC, ndrange = size(dS))`. Both kernels sum into a **local** `temp` and assign once — `dA[l, m, h] = temp` at `:30`, `dS[l, h] = temp` at `:49`. Same shape at `mat_tensor_mul.jl:174`–`182` (four buffers, four full-coverage launches, writes at `:100`,`:120`,`:218`,`:241`), `tensor_mat_skew_sym_assign.jl:47`–`53` (writes at `:15`,`:31`), `vec_tensor_mul.jl:30` (write at `:23`). And `grep -rnE '^\s*(dA\|dS\|dZ\|a_diff)[A-Za-z_]*\s*\.?\+=' src/kernels/kernel_ad_routines/` exits 1 — nothing accumulates into any of them. |
| 2 | nit | `CHANGELOG.md` §9 / PR body §9 | "Each of those kernels was read first: all five write every `C[i, j, l]`" — the count is wrong under either reading. | Ten wrappers changed: six `zeros` → `allocate`, four `copy`/`zero` → `similar`. The four structured kernels write `C[i, j, l]`; the six generic ones write `C[i, j, k]`. The substance holds — I read all ten, listed below. |
| 3 | quality | `src/layers/multi_head_attention.jl:123` | The **matrix** method does not need the type assertion. `reduce(vcat, head_outputs)` takes Base's linear `AbstractVecOrMat` path — which the PR body itself states — infers concretely without an assertion, and measures cheaper. | `Base.infer_return_type(v -> reduce(vcat, v), Tuple{Vector{Matrix{Float32}}})` is `Matrix{Float32}`. Allocations and time, minimum over 20/2000 runs: 8 heads of 8×32 — `reduce` 8,272 B / 0.58 µs against splat-plus-assert 8,464 B / 0.92 µs; 32 heads of 4×64 — 32,848 B / 2.33 µs against 33,424 B / 3.21 µs. Results identical. The **tensor** method genuinely needs the variadic form; only this one line is affected. |
| 4 | nit | `CHANGELOG.md` §9 | The two dead-code removals the PR body lists — the `time_indices` comprehension in `convert_input_and_batch_indices_to_array` and the duplicated `size(output_diff)` destructure in `augment_zeros` — appear in the PR body but in no CHANGELOG entry. | `git diff origin/main...origin/pr/297 -- CHANGELOG.md` has three `### Changed` bullets and one `### Fixed` bullet; neither removal is named in any of them. |

On finding 3, for the record: the assertion is *not* generically safe. `vcat(v...)::eltype(v)` raises
a `TypeError` for any head-output type whose `vcat` falls back to `Array`. I measured `SubArray`,
`Adjoint` and a plain user `AbstractMatrix` subtype all failing `r isa eltype(v)`, against `Array`,
`BitArray` and a `Vector{Any}` of matrices passing. Nothing in this package reaches that case, which
is why this is a quality point and not a bug — see the verified list.

### PR-body claims: verified

All measurements in cold, separate `julia --startup-file=no --project=. --check-bounds=auto`
processes, Julia 1.13.0, arm64 macOS. `origin/main` and `origin/pr/297` never shared a process.

**§8, the minibatch index set.** The table reproduces to the byte, in both columns:

| columns | before, mine | before, PR | after, mine | after, PR |
|--:|--:|--:|--:|--:|
| 200 | 13,904 B | 13,904 B | 10,464 B | 10,464 B |
| 2,000 | 369,824 B | 369,824 B | 99,440 B | 99,440 B |
| 8,000 | 4,530,736 B | 4,530,736 B | 397,088 B | 397,088 B |

The wall-clock column does not transfer — I get 7.82 ms where the PR reports 41.77 ms at 8,000
columns — but the *ratio* the CHANGELOG draws its conclusion from does: 7.822/0.480 = 16.3× for four
times the minibatches, against the "fifteen times" in the text, and 0.031/0.009 = 3.4× after. The
sentence stands; the absolute milliseconds are machine-specific.

`Base.infer_return_type(batch_over_two_axes, Tuple{Batch, Int, Int, DataLoader})` returns
`Vector{Vector{Tuple{Int64, Int64}}}`, and the new test's ratio comes out at 3.993 against its bound
of 6 — comfortable. `Base.infer_return_type` exists from 1.11 and `[compat] julia = "1.11"`, so
`Julia min` can run the test; on 1.10 it does not exist, which is only worth noting if the floor
ever drops back.

**The caller-visible return-type change is genuinely covered.** Both `(batch::Batch)(dl)` methods
route through `batch_over_two_axes` (`src/data_loader/batch.jl:183`, `:191`), so no path still
returns a tuple. The only consumer in `src/` is the `for batch_indices in batches` loop at
`src/data_loader/optimize.jl:69`, which is unaffected. Every other use is `length`, `length.`,
`union(…)` or `batches[i]` — identical for a vector. The four `length.` assertions in the two
docstrings, the four in `test/docstrings/data_loader.jl`, the prose line in each docstring, and the
math block plus the "vector of vectors" sentence in `docs/src/data_loader/data_loader.md` are all
updated. `grep -rni "tuple of vector"` over `src/`, `test/` and `docs/` returns nothing. The
`@example batches` block at `data_loader.md:77`–`90` indexes `batches[1]`/`batches[2]`, which is
fine either way, and `Doctests` is green on this head.

**§9, `compute_output_of_mha`.** Every row reproduces exactly, and the output checksums are
identical between `main` and the branch at all three sizes:

| dim, heads, len, data | path | before, mine = PR | after, mine = PR |
|:--|:--|--:|--:|
| 64, 8, 32, 16 | matrix | 176,592 B | 150,272 B |
| 64, 8, 32, 16 | tensor | 2,732,208 B | 2,261,296 B |
| 128, 16, 64, 16 | matrix | 1,221,424 B | 975,168 B |
| 128, 16, 64, 16 | tensor | 19,308,848 B | 15,306,992 B |
| 128, 32, 64, 16 | matrix | 2,311,760 B | 1,802,688 B |
| 128, 32, 64, 16 | tensor | 36,233,776 B | 28,008,560 B |

The 4-head regression is real and exactly as stated: 6,352 → 6,944 B on an 8×6 matrix. The tensor
path at 4 heads *improves*, 21,488 → 19,360 B, which the body does not claim either way.

**The type assertion is correct for everything this package can produce.** I exercised
`compute_output_of_mha` at `(dim, heads)` of (64, 8), (8, 4) and (8, 1), with `Stiefel = true` and
`false`, on matrix and tensor input, plus a `SubArray` matrix, an `Adjoint` matrix and a `SubArray`
tensor. All return `Matrix{Float32}` or `Array{Float32, 3}`; none throws. The reason is structural
rather than lucky: on the tensor side every head output comes from `KernelAbstractions.allocate`,
and on the matrix side `*` materialises a dense `Array` whatever the input wrapper is. Inference on
the branch gives `Matrix{Float32}` and `Array{Float32, 3}` — which is what `main` gives, so the
assertion restores exactly what splatting lost, as claimed.

**The head key change is safe.** `Symbol("head_", 3) === Symbol("head_" * string(3))` is `true`, and
the only consumers outside the layer index it by field name:
`test/transformers/transformer_setup.jl:9`, `test/transformers/transformer_optimizer.jl:36`–`37`.

**§9, kernel buffers — I read all ten.** Each writes `C` unconditionally over `ndrange = size(C)`,
so nothing reads an uninitialised element:

- `zeros` → `allocate`: `mat_tensor_mul_kernel!` (`src/kernels/mat_tensor_mul.jl:11`),
  `tensor_mat_mul_kernel!` (`tensor_mat_mul.jl:11`), `tensor_tensor_mul_kernel!`
  (`tensor_tensor_mul.jl:11`), `tensor_tensor_transpose_mul_kernel!`
  (`tensor_tensor_transpose_mul.jl:15`), `tensor_transpose_mat_mul_kernel!`
  (`tensor_transpose_mat_mul.jl:14`), `tensor_transpose_tensor_mul_kernel!`
  (`tensor_transpose_tensor_mul.jl:15`).
- `copy`/`zero` → `similar`, the four that matter, since an unwritten element would be visible
  there: `symmetric_mat_mul_kernel!` (`mat_tensor_mul.jl:88`), `lo_mul_kernel!` (`:132`),
  `up_mul_kernel!` (`:179`), `skew_mat_mul_kernel!` (`:229`). Each assigns `C[i, j, l] = tmp_sum` on
  every index of `ndrange = size(C)`, and `C = similar(B)` gives `size(C) == size(B)`.

**`augment_zeros` really is different**, so that exclusion is right:
`src/data_loader/tensor_assign.jl:88`–`89` launches with `ndrange = size(output_diff)` while
`zero_tensor` is `sys_dim × seq_length × batch_size` — strictly larger in the second axis. The zeros
are load-bearing. The removed duplicate destructure only re-bound `prediction_window` and
`batch_size` to the same values and bound an unused `dim`; the surviving line at `:84` is the only
one read.

**The isolated micro-benchmarks reproduce**, KernelAbstractions 0.9.42, 64³ `Float32`: `KA.zeros`
28.8 µs (PR: 31.3), `KA.allocate` 1.12 µs (1.08), `copy` 11.3 µs (12.4), `zero` 8.4 µs (8.5),
`similar` 0.041 µs (0.04). The 12,400-byte difference is exact: `KA.zeros` allocates 1,061,440 B
against `allocate`'s 1,049,040 B, for a 1,048,576-byte array.

**§10, the minibatch copies — including the case not on the PR's list.** Five of the six
`convert_input_and_batch_indices_to_array` methods return freshly `allocate`d arrays, so `_copy` was
pure waste there. The sixth, `src/data_loader/batch.jl:376`, returns `@views` — that is where
`_copy` did real work, and it is not among the networks the body lists. It is reachable:
`DataLoader(input::AbstractArray{T, 3}, output::AbstractArray{T, 3})` at
`src/data_loader/data_loader.jl:179` shares `T` but not the array *type*, so passing a view as the
output selects it over the `{T, BT, BT}` method at `batch.jl:387`. I built that loader and trained
through it. It works, and it is cheaper without the copies, not dearer:

| path | main | branch |
|:--|--:|--:|
| dense output (`batch.jl:387`), 5 epochs | 6,457,568 B, 2.79 ms | 5,974,768 B, 2.14 ms |
| view output (`batch.jl:376`), 5 epochs | 22,401,248 B, 7.75 ms | 21,503,728 B, 6.81 ms |

So the deletion holds on the one path the body's evidence does not cover. `grep -rn "_copy"` finds
no surviving reference to the deleted helper in `src/`, `test/` or `docs/` — the remaining hits are
`deepcopy`, `ps_copy` locals and `GeometricOptimizers._copyto!`. `QPT` and `QPTOAT` are still used
in `src/data_loader/data_loader.jl`, so removing the three `_copy` methods orphaned no alias.
`time_indices` survives only in `batch_over_two_axes`, read at `batch.jl:168`.

**The `reduced_system.jl` comment is accurate.** `decoded_nt = decoder((q = q̃, p = p̃))` at `:115`
and `ForwardDiff.jacobian(qp -> decoder(qp), vcat(q̃, p̃))` at `:118` are two evaluations, as the
comment says; it is present tense and asserts only what is visible. C14 sits under *C. Follow-ups
and cleanups*, and `test/reduced_order_modeling/reduced_system.jl` does exercise `HRedSys`, so the
entry's "the change is checkable" holds.

### Ordinary pass

- **Scope.** All 16 files trace to §8, §9 or §10. Nothing outside.
- **Piracy.** The diff defines no method on a type owned elsewhere. `vcat` is called, never extended.
- **Imports.** No `using`/`import` line changed, and no removal orphaned one.
- **Formatting and NFC.** JuliaFormatter under the repo's `.JuliaFormatter.toml` reports all
  fourteen changed `.jl` files clean; `nfc.jl` exits 0 on the changed sources, `CHANGELOG.md` and the
  docs page.
- **Comment tense and history.** Present tense throughout. The one to watch is the header of
  `test/data_loader/batch_index_set.jl`, which describes the splatted-tuple anti-pattern the test
  guards against. It reads as the rationale for the bound rather than as package history, so I would
  leave it.
- **Stale `file:line` citations.** None. A grep for `<changed file>:<line>` across `src/`, `test/`,
  `docs/` and `CHANGELOG.md` returns nothing, so this diff falsified no witness.
- **CHANGELOG placement.** Entries sit under `## [Unreleased] — 0.8.0`, split correctly between
  *Changed* (behaviour) and *Fixed* (the index set), with C14 under *C. Follow-ups and cleanups*.
- The `@test_broken` the body attributes the two broken tests to is real, at
  `test/parameters/network_parameters_gradient_structure.jl:41`.

### Not checked

- **The full `Pkg.test()` figures** — 74 testsets, 4302 passing, 2 broken, 0 failed. Left to CI
  rather than re-run locally.
- **The `MetalBackend()` run.** No GPU probe from this session. The CPU reasoning above says the
  assertion cannot throw for any `vcat` that returns the backend's own array type, which matches
  what the body reports, but I did not execute it.
- **`Documenter.doctest` locally.** The `Doctests` job is green on this head, which is the stronger
  evidence anyway.
- **Aqua and ExplicitImports.** Not re-run: the diff adds no method on a foreign type and changes no
  import, so neither result can move.

### CI at the time of writing

Run `35376717538` on `d26c612` is still going. Green so far: `Julia 1 - ubuntu-latest` (26m38s),
`Julia 1 - macOS-latest` (19m55s) and `Doctests - ubuntu-latest` (1m40s). Still running: both
`Julia min` jobs, `Julia 1 - windows-latest`, `Julia nightly`, `Julia pre`, `Scripts` and
`Documentation`. Nothing has failed. The two advisory jobs (`pre`, `nightly`) are
`continue-on-error` and are not required checks.

michakraus:
**CI update**, correcting the "nothing has failed" line in my review above: run `35376717538` has
finished, and every **required** check is green — `Julia 1` and `Julia min` on all three platforms,
`Doctests`, `Documentation`, and both codecov contexts. `Scripts - ubuntu-latest` is still pending.

`Julia nightly - ubuntu-latest` went **red** after 1h0m47s. That job is `experimental: true` +
`continue-on-error` and is not a required check, so it does not block. I could not read its log —
`gh run view --log-failed` and the check-run API both return HTTP 403 secondary rate limit right
now — so I cannot say whether it is a resolver failure, a nightly-only inference change or a real
test failure, and I have not compared it against nightly on `main`. Worth one look before merge;
the verdict on the diff itself is unchanged.

michakraus (commented):
## Review — Audit Part C

The three changes are sound and the measurements hold. I reproduced the allocation tables in cold
processes and every figure in both of them matches to the byte. Four findings, all now fixed in
`b6709acf` on this branch; the one code change is in the matrix path of `compute_output_of_mha`.

I wrote that commit, so the merge is not mine to make.

### 1. `CHANGELOG.md` — the `kernel_ad_routines` buffers are not accumulators

The entry justified leaving them alone with *"the `kernel_ad_routines` buffers are accumulators"*.
They are not. Every kernel under `src/kernels/kernel_ad_routines/` sums into a **local** and assigns
the buffer element once:

- `mat_tensor_mul.jl:30` `dA[l, m, h] = temp`, `:49` `dS[l, h] = temp`, and the same at `:100`,
  `:120`, `:218`, `:241`
- `tensor_mat_mul.jl:33`, `:54`
- `tensor_mat_skew_sym_assign.jl:15`, `:31`
- `vec_tensor_mul.jl:23`

Every launch uses `ndrange = size(buffer)` — `mat_tensor_mul.jl:67,68,138,139,179–182,258,259`,
`tensor_mat_mul.jl:71,72`, `tensor_mat_skew_sym_assign.jl:52,53`, `vec_tensor_mul.jl:33`. A grep for
an accumulating write on any of those buffers returns nothing; every `+=` in the directory is on a
local `temp` or `a_val`.

So they have exactly the shape §9 optimised, and the backward pass is where training spends its
time. The sentence is replaced with the real reason — this part was scoped to the forward path — and
the buffers are catalogued as **C15**.

The `augment_zeros` exclusion beside it is genuine and stays: `zero_tensor` is
`sys_dim × seq_length × batch_size` while the launch is `ndrange = size(output_diff)` with
`prediction_window` in axis 2, so the kernel writes a slice and the zeroing carries load.

### 2. `CHANGELOG.md` — "all five write every `C[i, j, l]`"

Ten wrappers changed, not five: six `zeros` → `allocate` and four `copy`/`zero` → `similar`. Six of
them index `C[i, j, k]` and four `C[i, j, l]`. The substance is right — I read all ten, and each
launches over `ndrange = size(C)` and writes every element — but the count and the index are wrong.
Corrected.

### 3. `src/layers/multi_head_attention.jl` — the matrix method needs no assertion

`reduce(vcat, …)` takes Base's linear path for `AbstractVecOrMat`, which the entry says itself when
it explains why the tensor method cannot use it. On the matrix method it infers concretely with no
assertion and allocates slightly less, so the splat buys nothing there. The matrix method now reads

```julia
reduce(vcat, head_outputs)
```

The tensor method is untouched: `reduce` folds pairwise on a 3-tensor, so the splat and its
assertion stay.

Matrix path, cold `--check-bounds=auto` processes, the PR head against this change. Checksums and
Zygote gradients are identical at every size, and the tensor path is unchanged to the byte:

| dimension, heads, length, data | PR head | `reduce` |
|:--|--:|--:|
| 64, 8, 32, 16 | 150,272 B | 150,096 B |
| 128, 16, 64, 16 | 975,168 B | 974,864 B |
| 128, 32, 64, 16 | 1,802,688 B | 1,802,128 B |
| 8, 4, 6, 4 | 6,944 B | 6,832 B |

The 4-head regression the entry records shrinks from +592 B to +480 B against `main`. The
CHANGELOG table and that sentence are updated.

**The assertion on the tensor method is load-bearing, and I checked rather than assumed.** With it
removed, `compute_output_of_mha` infers as `Any` for tensor input and the `Any` reaches the functor.
Keeping it is right.

### 4. `CHANGELOG.md` — two removals were in no entry

The `time_indices` comprehension in `convert_input_and_batch_indices_to_array` and the duplicated
`size(output_diff)` destructure in `augment_zeros` are in the PR description but in none of the
bullets. Both are in the diff. A bullet now records them.

### Added: an inference test

The description says the pre-PR gate caught the inference regression and the suite did not, because
nothing in `test/` asserts the inference of a forward pass. Removing the matrix assertion makes that
gap load-bearing, so `test/attention/multi_head_attention_inference.jl` pins the inferred return
type of both `compute_output_of_mha` methods and of the functor, for `Stiefel = false` and
`Stiefel = true`. All eight cases pass.

### Verified from the description

- **Both allocation tables reproduce to the byte**, in cold processes, baseline and branch measured
  separately. The §8 wall-clock figures do not transfer to this machine — 7.82 ms against 41.77 ms
  at 8,000 columns — but the ratio the conclusion rests on does: 16.3× before, 3.4× after.
- `batch_over_two_axes` returns `Vector{Vector{Tuple{Int64, Int64}}}`. The new test's ratio is 3.993
  against its bound of 6. `Base.infer_return_type` exists from 1.11 and `[compat] julia = "1.11"`,
  so `Julia min` can run it.
- **The caller-visible change is fully covered.** Both functor methods route through
  `batch_over_two_axes`; the only `src/` consumer is the loop in `optimize.jl`. Everything else
  takes `length` or `union`. The docstrings, `docs/src/data_loader/data_loader.md` and
  `test/docstrings/data_loader.jl` all agree.
- **The assertion holds for everything the package produces** — twelve configurations across
  `Stiefel` and both input ranks, plus `SubArray` and `Adjoint` matrix inputs. None throws.
- **§10 is right, and reaches further than the description claims.** Five of the six `convert_…`
  methods return fresh arrays. The sixth returns `@views` and **is** reachable, through
  `DataLoader(::AbstractArray{T,3}, ::AbstractArray{T,3})`, whose two arrays share `T` but not their
  type. Training through that case: 22,401,248 B on `main` against 21,503,728 B here. The deletion
  wins there too. Nothing references `_copy` or `time_indices` any more.
- No type piracy, no import change, scope clean, comment tense right, no stale `file:line` citation.

### Not verified

- The `MetalBackend()` run in the description. No GPU probe from this session.
- The `Julia nightly` job. It is `experimental: true` and not required, and the GitHub API refused
  its log with a secondary rate limit, so I cannot classify it or compare it against nightly on
  `main`.

### CI

Every required check was green on `d26c612`: `Julia 1` and `Julia min` on ubuntu, macOS and
windows, `Doctests`, `Documentation`, and both codecov checks.

### Verification of the fixes

`Pkg.test()` on Julia 1.13 with `--check-bounds=auto`: **75 testsets, 4324 passing, 2 broken, 0
failed**, counted from the summary rows rather than from a grand total. The 2 broken are the same
pre-existing `@test_broken` the description names. JuliaFormatter and `nfc.jl` are clean on all four
files, and the pre-commit hook passed formatting, `fatou lint`, NFC and the load test.

That is 22 passes above the description's 4302, where the new file adds only 8. The rest arrived
with the merge commit: `ccbea026..d26c612` brings 58 lines into
`test/transformers/positional_encoding.jl` from #296. The description's figures were measured before
that merge, so the table wants re-running either way.

PR #299, opened 2026-09-18
Remove the dead code under src/, and gate the imports against their return

michakraus:
Part E of the `GeometricMachineLearning` audit — sections §14–§16. It removes dead code from `src/`, clears thirteen stale explicit imports, and adds a gate so that the imports cannot come back unobserved.

`src/` goes from **82 files and 9,022 lines to 76 and 8,732**. Eighty-nine of those 290 lines moved rather than went, with the kernel generator.

## Breaking

`UnknownProblem` and `NothingFunction` are no longer exported, and the types are gone with the export. They stood in the module file under the comment `# are these needed?`. The answer is no: the comment's own question had two occurrences in the repository — the definition and the export. `NothingFunction`'s only reader was `is_NothingFunction`, which nothing called either.

**The thirteen removed imports break a second thing.** An `import` in the module file makes a name resolve as `GeometricMachineLearning.<name>` whether or not the package exports it, so dropping the thirteen closes thirteen such paths. Two had users in this repository and are repaired below; the other eleven had none here, so an outside caller is the only one that can notice. Dispatch is unchanged: no removed definition extends a foreign generic, so no surviving name resolves to a different method.

## §14 — five files and one function

| file | what it held |
|:--|:--|
| `kernels/vector_matrix.jl` | `vec_add!`, 5 lines, never `include`d — and written in **CUDA.jl** syntax (`threadIdx().x`) for a package with no CUDA dependency |
| `kernels/tensor_transpose_tensor_transpose_mul.jl` | a closed cluster: kernel ← wrapper ← constructor, whose only other occurrence was the `include` string |
| `data_loader/matrix_assign.jl` | `draw_batch!`, uncalled — and carrying a latent index-0 bug, `Int.(ceil.(rand(T, n) * n_params))`, which fires exactly 1 draw in 2,048 at `Float16` |
| `architectures/fixed_width_network.jl` | `struct FixedWidthNetwork <: Architecture end`, and nothing else |
| `architectures/variable_width_network.jl` | `struct VariableWidthNetwork <: Architecture end`, and nothing else |

`tensor_exponential` goes but its file stays, because `init_output` below it is live — `tensor_cayley.jl` calls it five times and `cpu_inverse.jl` once — and `init_output` is in turn the only caller of `assign_ones!` next to it. Removing `tensor_exponential` also retires an unbounded `while true`: the Taylor series had no iteration cap, so an input converging more slowly than `eps(T)` would not have terminated.

The kernel generator moves to `scripts/utilities/inverse_generator.jl`. It is outside `verification/` and `reproduction/`, so the script gate does not run it — which is right, since it needs `Symbolics` and is run by hand.

## §15 — the uncalled definitions

`src/utils.jl` loses 63 of its 167 lines. Elsewhere: two never-launched kernels in `data_loader/tensor_assign.jl`, the `AbstractCache` backward-compat alias, `default_retr` (also a non-`const`, untyped module global), `SymplecticDimensionChange`, `ClassificationTransformerLoss` and its functor, the deprecated `Gradient` constructor that `@warn`ed on every call, and `∇L`, `∇∇L`, `∇q̇∇q̇L`.

**The three Lagrangian derivatives are the one judgement call.** A comment described them as the reference for what `LNNLoss` computes by a different route, "kept for checking that route by hand". No test ever took the comparison and nothing called them, so what stood there was an intention rather than a check. Restoring it wants a test; the definitions are in the history either way.

## §16 — the imports

Thirteen stale explicit imports removed, and `StateVariable` now comes from `GeometricBase`, which defines it, rather than `GeometricSolutions`, which re-exports it. `fatou lint`'s `unused-import` count on the module file drops **40 → 27**, which is the thirteen and nothing else.

This is not tidiness. A live-but-useless `using` keeps a dependency reachable from the module file, so Aqua's `stale_deps` reads it as in use. That check passes today and only means something now that this list is empty.

## The gate, and what it does not gate

`test/aqua.jl` gains a `test_explicit_imports` call running three of seven checks: `no_stale_explicit_imports`, `all_explicit_imports_via_owners`, `no_self_qualified_accesses`. The first two were red before. The other four are off, each with its count and its reason written above the call — 37 implicit imports, 11 non-public explicit imports, 2 non-owner qualified accesses, 22 non-public qualified accesses — and recorded as **C16** under `## Open Issues`.

**This closes the import half only.** `ExplicitImports` reads imports, never definitions, so re-adding any of the thirty-two removed definitions leaves the suite green. Nothing here gates a definition, and nothing cheap could: separating a dead definition from a live one is the dispatch question that took two sweeps.

`all_explicit_imports_via_owners` also carries the property the `ambiguities` comment in the same file rejects. It asks *which* upstream module defines a name, so it can go red when a name moves between `GeometricBase`, `GeometricOptimizers` and `AbstractNeuralNetworks` with nothing in `src/` having changed — which is exactly how `StateVariable` got here. `ExplicitImports = "1.15"` admits any 1.x, so a minor release that sharpens stale detection does the same. It stays a gate anyway, because unlike an ambiguity count a red here names the import and the module to take it from, and the fix is one line. The comment block says so.

`test/data_loader/accuracy.jl` is not a gate on this removal. It exercises `accuracy`, which survives it — a function `docs/src/data_loader/data_loader.md` renders and nothing called. The two assertions are exact and seed-independent: label every sample with the class the untrained network already predicts and the accuracy is 1; rotate every label one class on and it is 0.

## What this exposed

**`ExplicitImports` sees `src/` only, and ten occurrences outside it broke.** `layer` and `Ω` are reported stale because nothing in `src/` uses them, and that is true. What the tool cannot see is that `test/`, and one script, reached them *through the package* — as `GeometricMachineLearning.layer` and `GeometricMachineLearning.Ω`.

| file | name | occurrences | fixed to |
|:--|:--|--:|:--|
| `test/layers/sympnet_upscaling.jl` | `layer` | 4 | `using AbstractNeuralNetworks: layer` |
| `scripts/verification/sympnet_upscaling_symplecticity.jl` | `layer` | 4 | the same |
| `test/docstrings/manifolds.jl` | `Ω` | 2 | `using GeometricOptimizers: Ω` |

Reaching the generic where it is defined is the right fix in all three files, and no import was restored. The first full `Pkg.test()` was red on this, and neither `ExplicitImports` nor the package loading nor the two test guards saw it coming.

**`accuracy` cannot be reached through any public `DataLoader` constructor.** Its signature needs a `DataLoader` whose input and output element types differ — real inputs, integer one-hot targets — and every `DataLoader(input, output)` method binds both arrays to one `T`. The test therefore names the parametric type. Recorded as **C17**; closing it is a two-element-type constructor, which is an API addition rather than a cleanup.

## Verification

Full `Pkg.test()` on Julia 1.13 at `--check-bounds=auto`: **76 testsets, 4,331 passing, 0 failures, 0 errors, 2 broken**. The two broken are pre-existing (`_custom_mul` and `Symplectic attention`, both `NetworkParameters gradient structure`).

**Fixed during the pre-PR pass**
- `src/optimizers/optimizer.jl` — the `AbstractCache` note said its export went in 0.7.0. It went in 0.5.0: `export OptimizerMethod, AbstractCache` is present at `v0.4.8` and absent at `v0.5.0`. The release number is dropped rather than corrected, since version history belongs in `CHANGELOG.md`.
- `test/aqua.jl` and `CHANGELOG.md` — `no_implicit_imports` was described as 37 names across twelve packages. `explicit_imports` returns 37 rows across **thirteen** modules.
- Nine further corrections in `CHANGELOG.md`: five `file:line` citations this branch shifted (`src/utils.jl:53`→`:21`, `GeometricMachineLearning.jl:90`→`:84`, `:189`→`:180`, `optimizers/optimizer.jl:21`→`:22`), three counts (`src/utils.jl` is 167 lines not 168, so 38 % not two thirds; `1 in 1,980` is exactly `1 in 2,048`, since `rand(Float16)` is uniform over the 2,048 multiples of `2^-11` in `[0, 1)`), and three `[Unreleased]` passages that said the Lagrangian derivatives, the `rdevelop`/`develop` methods and `_add` were kept, which this branch removes. `[Unreleased]` is read as one release, so those had to be reconciled rather than left to contradict each other.

**Fixed after review**
- `CHANGELOG.md` and this description said "nothing else here changes a name a caller can reach". That is false: the thirteen removed imports close thirteen qualified paths, ten occurrences of which this branch repairs itself. Both now say so.
- The title and both gate descriptions claimed a gate against the dead code's return. The gate covers the imports only. Both now say what it does not catch, and `test/aqua.jl` carries the same caveat plus the upstream-move failure mode of `all_explicit_imports_via_owners`.
- `CHANGELOG.md` said `tensor_cayley.jl` and `cpu_inverse.jl` call `assign_ones!`. They call `init_output`; `assign_ones!` is reached only from `init_output`.
- The commit message of `543ab15` says the imports broke nine call sites. It is ten — four, four and two across three files. The count is corrected here and in `CHANGELOG.md`; the pushed commit message is left as written rather than force-pushed over.
- The count "26 deleted names" below was stale. It is 32, as `CHANGELOG.md` states: 25 from surviving files and 7 with the five deleted ones.
- `test/aqua.jl` — the wrapping `@testset "ExplicitImports"` is removed. `test_explicit_imports` opens its own testset of that name, so the report nested `ExplicitImports > ExplicitImports`.

**Checked and clean**
- Every one of the 32 deleted names grepped over `src/`, `test/`, `docs/`, `scripts/`, `ext/` and the `.toml` files. The only live hits are this branch's own explanatory comments. No `@docs` block or `@ref` in `docs/src/` names a removed symbol.
- Type piracy unchanged at 12, asserted by `test/aqua.jl` under a passing suite. No changed hunk defines a method on a foreign generic.
- Ambiguities measured in a fresh process: 18 with `GeometricMachineLearning` alone, 17 of them `PoissonTensor` — unchanged by removing `ClassificationTransformerLoss <: NetworkLoss`, so the comment in `test/aqua.jl` still holds.
- Type instability and allocations as regression gates: no live method body changes. The one non-deletion in `src/` is the `StateVariable` import move, and `GeometricBase.StateVariable === GeometricSolutions.StateVariable` is `true`.
- JuliaFormatter and `nfc.jl` clean over every changed file.

**Not checked**
- The Documenter build. `docs/` is untouched, no docstring was deleted or detached, and no `@docs`/`@ref` names a removed symbol — but only a build proves it. The `Documentation` job on this PR is the check, and it passed.

**Pre-existing, not fixed here**
- `src/data_loader/tensor_assign.jl:2` keeps a history note in its docstring ("developed during Cemracs"), directly above a changed hunk. That belongs to Part F.
- `Julia nightly` is red — Aqua `unbound_args` over one method, `src/architectures/neural_network_integrator.jl:98`, a file this branch does not touch. `main` fails identically. Recorded as **B9** under `## Open Issues`.

## Notes

§13 was not folded in, so the `add!` family is untouched and Part D still carries §11, §12 and §13.


michakraus (commented):
## Review

Reviewed at `543ab15`. The removal is sound, and nothing found blocks it. All six findings are in what the branch *says* about itself, not in the code it deletes. They are fixed in `cb9950d8`.

### Why the deadness claim holds

A grep is normally not evidence for a dead-code claim, because a method added to an existing generic, or a qualified definition such as `Base.similar(...)`, is invisible to a text search. Here it is evidence, and the reason is checkable: `main`'s module file imports **none** of the thirty-two removed names, and every removed definition in the diff is an unqualified `function`, `struct`, `const` or `@kernel` in this package's own namespace. No removed definition extends a foreign generic, so no dispatch question is left open. `grep -rn "subtypes(" src/ test/ docs/ scripts/ ext/` returns nothing, so no subtype enumeration depended on `ClassificationTransformerLoss`, `FixedWidthNetwork` or `VariableWidthNetwork`. `ReducedComplexityModeling`, `MultiSymplectic` and `SymbolicNeuralNetworks` name none of the thirty-two and none of the thirteen imports.

`Gradient` is worth naming separately: `main` never imports it — `src/optimizers/optimizer.jl` reaches `GeometricOptimizers.Gradient` qualified — so the removed deprecated constructor was not a method on a foreign generic either.

### Findings

| # | severity | file:line | finding |
|--:|:--|:--|:--|
| 1 | quality | `CHANGELOG.md:54`, description | "Nothing else here changes a name a caller can reach" is false. An `import` in the module file makes a name resolve as `GeometricMachineLearning.<name>` whether or not the package exports it, so the thirteen removed imports close thirteen qualified paths. The branch repairs ten occurrences of two of them. The other eleven had no user here. The release is already breaking, so the fix is the sentence. |
| 2 | quality | title, `CHANGELOG.md:1262` | "Gate against its return" covers the import half only. `ExplicitImports` reads imports, not definitions, so re-adding any of the thirty-two removed definitions leaves the suite green. The `accuracy` test exercises a function that survives the removal and gates nothing in it. |
| 3 | risk | `test/aqua.jl:105` | `all_explicit_imports_via_owners` has the property the `ambiguities` comment at `test/aqua.jl:66` rejects: it asks *which* module owns a name, so it goes red when a name moves between `GeometricBase`, `GeometricOptimizers` and `AbstractNeuralNetworks` with nothing in `src/` changed — which is how `StateVariable` got here. `ExplicitImports = "1.15"` admits any 1.x, so a minor release that sharpens stale detection does the same. It should stay on: unlike an ambiguity count, a red names the import and the module to take it from. It wants the caveat written down. |
| 4 | nit | `test/aqua.jl:105` | The wrapping `@testset "ExplicitImports"` duplicates one `test_explicit_imports` opens itself, so the report nests `ExplicitImports > ExplicitImports`. Unlike the `Aqua` block, this wrapper carries no second assertion that needs a parent. |
| 5 | nit | commit message `543ab15` | "broke nine call sites" — it is ten: four `layer` in `test/layers/sympnet_upscaling.jl`, four in `scripts/verification/sympnet_upscaling_symplecticity.jl`, two `Ω` in `test/docstrings/manifolds.jl`. |
| 6 | nit | `CHANGELOG.md:67` | "`init_output` and `assign_ones!` … `tensor_cayley.jl` and `cpu_inverse.jl` call them" — those two files call `init_output` only, six times between them. `assign_ones!` is reached only from `init_output`. Live, but transitively. |

The description also said 26 deleted names where the changelog says 32. The changelog is right.

### Fixed in `cb9950d8`

All six. `CHANGELOG.md` and the description now state the qualified-path break and the ten repaired occurrences; both, the title and the comment block in `test/aqua.jl` now state what the gate does not catch and how `all_explicit_imports_via_owners` can go red on an upstream move; the wrapper testset is gone; the two miscounts are corrected. The commit message of `543ab15` is left as written rather than force-pushed over, with the corrected count recorded in the changelog and the description.

`test/aqua.jl` was re-run in the test environment after the wrapper came off: the bare `test_explicit_imports` call does run inside the `@safetestset` that `runtests.jl` wraps the file in, and the three checks pass. 13 assertions, all passing, piracy still exactly 12. JuliaFormatter and the NFC check are clean on both changed files.

### CI at `543ab15`

Green. Seven of eight matrix jobs `success` (`min`/`1`/`pre` × ubuntu/macOS/windows), `Doctests`, `Documentation` and `Scripts` all `success`, both codecov contexts pass. The `Documentation` job passing is what proves no removal left a docstring detached or an `@ref` dangling — a failure of that kind passes format, lint, load and `Pkg.test()`.

`Julia nightly` is `failure` and it is **pre-existing**: Aqua `unbound_args` over one method, `var"#iterate#55"` at `src/architectures/neural_network_integrator.jl:98`, a file this branch does not touch. Run 35394536346 on `main` fails identically at the same line. Already recorded as **B9**.

### Verified

Every quoted figure was re-measured: the file and line counts, the 63 lines out of `src/utils.jl`, the 89-line generator, 32 = 25 + 7, the `fatou lint` `unused-import` drop of exactly 13, the `AbstractCache` export moving at 0.5.0, `StateVariable` being `GeometricBase`'s, and the `rand(Float16)` zero rate (2,027 zeros in 2²² draws, 1 in 2,069, spacing 2⁻¹¹). Every `file:line` citation the branch rewrote reads as stated. No `@docs` block or `@ref` under `docs/src/` names a removed symbol. `test/exports.jl` builds its list from `names(...)`, so dropping the two exports needs no edit there. The repaired call sites resolve: `AbstractNeuralNetworks` is in both `scripts/Project.toml` and `test/Project.toml`, `GeometricOptimizers` in `test/Project.toml`.

### Not checked

The four disabled `ExplicitImports` counts (37 / 11 / 2 / 22) are commentary rather than assertions and were not re-measured; nor were the ambiguity figures, which no gate depends on. The local `Pkg.test()` counts in the description were not reproduced — the green matrix stands in for the verdict, not for the numbers. `julia-callers.jl` was not run; the static argument at the top replaces it for these names.

### Verdict

Approve, with the corrections in `cb9950d8` folded in. Re-running CI on the new head is the remaining step.

PR #300, opened 2026-09-19
Close the audit's piracy and ambiguity findings (Part D)

michakraus:
Part D of the whole-package audit of v0.7.0: type piracy and ambiguities, closing **B7** (rewritten) and **B8** (closed).

| | before | after |
|:--|--:|--:|
| Aqua piracies | 12 | **3** |
| ambiguities, package alone | 18 | **1** |
| ambiguities, inside `Pkg.test()` | 27 | **1** |

`test/aqua.jl` now asserts both numbers, so neither can drift in either direction.

## Nine piracies deleted, because nothing called them

- `Base.:+(::Float64, ::Tuple{Float64})` and `Base.:+(::Vector{Float64}, ::Tuple{Float64})`. The second returned a **scalar** from a vector argument, silently discarding every element but the first. No call site in `src/`, `ext/`, `test/`, `docs/` or `scripts/`.
- `Base.:≈` on a `(q, p)` `NamedTuple` pair. One call site, in `test/arrays/poisson_tensor.jl`, which now compares the two fields. A caller who wants the old behaviour writes `a.q ≈ b.q && a.p ≈ b.p`.
- The six `add!` methods on `GeometricOptimizers`' structured matrix types, with `src/arrays/gml_extensions.jl`, its `include`, and the `import AbstractNeuralNetworks: add!` that made them piracy.

This **corrects B7**, which called four of the six load-bearing because the upstream generic is a `CanonicalIndexError` on those types. The reasoning is right and the conclusion is not: `add!` has no occurrence anywhere beyond the definitions and their own recursion, and `GeometricOptimizers` defines the identical set against its own generic.

It is under a breaking heading because `add!` **was** exported — `export _add, apply_toNT, add!` at v0.5.0 and v0.6.0 — and the export went earlier in this same release. At this branch's base it is no longer in `names(GeometricMachineLearning)`, which is what makes deleting the methods safe, but a 0.6.0 caller could reach them by name.

The three that remain are the `Dense`/`Linear` three-axis functors at `src/layers/resnet.jl:63,67,71`. Closing those needs an API change — upstream gaining the methods, or this package wrapping the two layer types — so B7 now covers only them.

## `PoissonTensor` no longer claims every argument type

`PoissonTensor <: AbstractMatrix{T}`, and two of its methods claimed an argument position wholesale, so they met every special-array method another package writes against `AbstractMatrix`:

- **`*` — 17 ambiguities** against `ArrayLayouts`, `FillArrays`, `Symbolics` and `GeometricOptimizers`. The three array methods now take `StridedVector`, `StridedMatrix` and `StridedArray{T, 3}`.
- **`getindex` — 9 more**, against the `Block`, `BlockIndex` and `BandRangeType` methods `BandedMatrices` and `BlockArrays` add to `AbstractMatrix`. Its indices are now `::Int`, which is the one method an `AbstractArray` must supply. Neither package loads with this one alone, so these nine were visible only from inside the suite — which is why the isolated and in-suite counts differed, and why they now agree.

The struct's second parameter is constrained to `AT <: AbstractMatrix{T}` at the same time.

The witness that closes: `PoissonTensor(4, Float32) * StiefelManifold(...)` was a `MethodError` between two names this package exports, and now resolves to `GeometricOptimizers`' own method.

**What a caller can notice.** A non-strided right-hand side reaches Julia's generic `AbstractMatrix` multiply instead of the specialised method, at the same value — only the fast path is given up, and for an argument no call site in this repository builds. An `Adjoint` or `Transpose` of a `Matrix` is in that class; a strided `view` is not. A 3-tensor has no generic fallback, so a non-strided one is a `MethodError`; nothing here passes one. The `(q, p)` `NamedTuple` method is untouched.

## Verification

`Pkg.test()` on Julia 1.13: **76 testsets, 4360 passing, 2 pre-existing broken, 0 failed, 0 errored.**

Every count above was read in a restarted session or a cold process. Revise does **not** retract deleted methods, so `Aqua.Piracy.hunt` kept reporting 12 in the session where the files had already gone, while the ambiguity count moved correctly — one number live and the other stale at the same moment.

The in-suite figures come from `julia --project=test`, which is the real test environment here because `test/Project.toml` carries a `[sources]` path to the package.

**Verified on an Apple GPU, not argued.** The `Strided…` bound is load-bearing for the GPU path: if `MtlArray` were not strided, `𝕁 * mtl_matrix` would fall through to the generic multiply, whose scalar `getindex` `GPUArraysCore` forbids — a silent break no CI job would see. `MtlMatrix <: DenseArray` holds, all three shapes satisfy the bounds, and vector, matrix and 3-tensor products stay on the device and agree with the CPU result at `maxdiff = 0.0`.

New assertions in `test/arrays/poisson_tensor.jl`: all four right-hand sides of `*` against the explicit `(q; p) -> (p; -q)` answer and against the plain matrix product; the non-strided fallback agreeing with the strided path; a strided `view` still taking the fast path; the `StiefelManifold` witness; and `getindex` for a scalar, a range, a `Colon`, a `CartesianIndex` and `end`.

## Pre-PR verification

Ten comment passages across three files narrated the deletion rather than describing the package, and were rewritten in the present tense or dropped where `CHANGELOG.md` already carries them.

Two errors of mine were caught and corrected, both in the justification rather than the decision:

- I wrote that the `add!` family "was never exported". It was, through v0.6.0. Corrected above.
- I listed an `Adjoint` of a `Matrix` among what `StridedArray` covers, in both the source comment and the changelog. It is not: `Adjoint{Float64, Matrix{Float64}} <: StridedMatrix` is `false`. It takes the generic path, at the same value.

Two further `[Unreleased]` passages contradicted this branch and are reconciled in place rather than deleted, since `[Unreleased]` is read as one release: the entry saying this package "only adds `add!` methods for the structured matrix types", and the `_GMLGradient` entry saying the remaining ambiguities are recorded under *Open Issues* and that `test/aqua.jl` "asserts neither".

Every `file:line` citation in the repository pointing into a changed file was read and is correct; four that this branch falsified are fixed.

## Pre-existing, not addressed here

- `test/aqua.jl` gives `all_qualified_accesses_are_public` as 22; the measured number is 24, drifted from upstream versions. No site is in a file this branch touches.
- `src/utils.jl:21-22` and one note in `test/aqua.jl` carry historical phrasing of the same class as those fixed here, outside this branch's hunks.
- The `fatou lint` `unused-import` findings on the module file are the known false class — that rule does not follow `include`, which is why `ExplicitImports` gates this instead.

## Note on #207

[#207](https://github.com/JuliaGNI/GeometricMachineLearning.jl/pull/207) raises the asserted piracy count from 12 to 17 and is currently `CONFLICTING` against `main`. Whichever of the two lands second must re-derive the number from `Aqua.Piracy.hunt(GeometricMachineLearning)` rather than resolving the conflict by hand, and update the B7 entry to match.


michakraus (commented):
## Review — Part D (piracy and ambiguities)

Reviewed from a detached worktree at `origin/pr/300` (`158f26d9`), against the merge base
`e5c96186`. Every number below was measured in a **cold process** against `--project=test`, which
resolves `GeometricMachineLearning v0.7.0` from the worktree through the `[sources]` path.

### The headline table reproduces, in both directions

Measured on this branch, and again on `origin/main` in a second worktree:

| | main (`e5c96186`) | PR head (`158f26d9`) |
|:--|--:|--:|
| `Aqua.Piracy.hunt` | 12 | **3** |
| `detect_ambiguities`, package alone | 18 | **1** |
| `detect_ambiguities`, after `using Zygote, GeometricIntegrators, HDF5` | 27 | **1** |

The in-suite process really does load the two extension packages that produced the nine extra
pairs (`BandedMatrices loaded: true`, `BlockArrays loaded: true`), so the second and third rows
are the two configurations the PR says they are. The three survivors are exactly the ones named:

```
(d::Dense{M, N, true})(x::AbstractArray{T, 3}, ps::NamedTuple)   src/layers/resnet.jl:63
(d::Dense{M, N, false})(x::AbstractArray{T, 3}, ps::NamedTuple)  src/layers/resnet.jl:67
(d::Linear{M, N})(x::AbstractArray{T, 3}, ps::NamedTuple)        src/layers/resnet.jl:71
```

and the single ambiguity is `resnet.jl:63` against `AbstractNeuralNetworks.Affine`'s functor, with
`Dense <: Affine` returning `false` — no common instance, so the pair is unreachable, as claimed.

### Findings

| 1 | correctness-risk | `src/arrays/poisson_tensor.jl:42-49`, `:81-96` | The narrowing does more than give up a fast path: it moves both generic paths onto **scalar `getindex` of the wrapped array**, which a GPU-backed `PoissonTensor` cannot serve. Neither the source comment nor the CHANGELOG says so, and the `getindex` half is not mentioned at all. | A `CountingMatrix <: AbstractMatrix` wrapper that counts scalar vs bulk `getindex`, used as the `J` field, run in a cold process on each branch. `𝕁[1:2, :]` — main: `scalar=0 bulk=1`; head: `scalar=8 bulk=0`. `𝕁[:, 1]` — main: `scalar=0 bulk=1`; head: `scalar=4 bulk=0`. `𝕁 * view(A, [1,2,3,4], :)` — main: `scalar=0`; head: `scalar=48`. `PoissonTensor(backend, n2, T)` (`:53`) is exported and documented at `:5` for exactly that case, and `GPUArraysCore` errors on scalar `getindex`. |
| 2 | quality | `test/aqua.jl:104` | The comment gives `all_qualified_accesses_are_public` as **22**; the measured number is **24**. The branch rewrote this very line (`the same class as the 11 above` → `the same class as the 10 above`), so the stale figure sits inside a hunk this branch changed. | `improper_qualified_accesses(GeometricMachineLearning, pathof(...))`, counting `public_access === false`, in a cold process on this branch: `NON-PUBLIC qualified accesses = 24`. Hunk `@@ -89,16 +92,16 @@` covers new lines 92–107. The neighbouring `10` at `:95` is correct — I measure `NON-PUBLIC explicit imports = 10 -> AbstractExplicitLayer, Architecture, _compute_loss, assign_columns, derivative, description, dim, load, model, save`. |
| 3 | quality | `src/utils.jl:21-22` | `# global_section(::AbstractVecOrMat) = nothing used to be defined here …` is a historical note outside `CHANGELOG.md`. It is pre-existing, but the hunk that deleted old lines 21–46 ends immediately above it, so it is now the first thing in the file after this branch's edit — and this branch is the pass that would take it. | Read at `src/utils.jl:21-22` on the head worktree. |

None of the three blocks the change. 1 and 2 are each a one-line edit.

### Verified and fine

- **The nine deleted methods have no caller.** `grep -rn "add!"` over `src/`, `ext/`, `test/`,
  `docs/`, `scripts/` returns only `GeometricOptimizers._add!` (a different name, at
  `src/optimizers/optimizer.jl:343`) and prose in `CHANGELOG.md`. Same for `_add`, `apply_toNT`,
  `AbstractTriangular` and `Tuple{Float64}` — no occurrence anywhere.
- **The claim of absence holds against the artefact, not the plan.** With the package loaded,
  `methods(AbstractNeuralNetworks.add!)` is 3 methods, all owned by `AbstractNeuralNetworks`;
  `count(m -> m.module === GeometricMachineLearning, ms) == 0`.
- **`GeometricOptimizers` really does carry the identical set.** Six `add!` methods against its own
  generic, at `skew_symmetric.jl:129`, `symmetric.jl:151`, `triangular.jl:16`,
  `stiefel_lie_algebra_horizontal.jl:117`, `grassmann_lie_algebra_horizontal.jl:116` and
  `utils.jl:51` in `GeometricOptimizers 0.7.0`, including the `AbstractVecOrMat` base case. And
  `AbstractNeuralNetworks` calls only the **two**-argument form, at `layers/dense.jl:21` and
  `layers/affine.jl:8`, on plain arrays — so nothing upstream reaches the deleted methods.
- **No new method is itself piracy.** The only addition is the `AT <: AbstractMatrix{T}` bound, and
  the piracy count went 12 → 3 rather than 12 → 3 with a swap.
- **The row-vector case is not disturbed.** The change narrows the left-`PoissonTensor` methods; it
  adds no `*(::AbstractMatrix, ::PoissonTensor)`. `v' * 𝕁 == v' * Matrix(𝕁)`, `v' * 𝕁 * v`,
  `𝕁' * v`, `transpose(𝕁) * v`, `𝕁 * 𝕁`, `𝕁 * I` all hold, and `v' * 𝕁` still returns an
  `Adjoint{Float64, Vector{Float64}}`.
- **The fall-through is value-correct.** `𝕁 * B'` (`Adjoint isa StridedMatrix` → `false`),
  `𝕁 * Diagonal(...)`, `𝕁 * BitMatrix` and `𝕁 * view(A, [1,2,3,4], :)` each equal the plain
  matrix product. `𝕁[2]` (linear), `𝕁[[1,2],[3,4]]`, `𝕁[:, 1]`, `𝕁[CartesianIndex(3,1)]` and
  `𝕁[end, end]` all still answer — so the narrowed `getindex` falls through to `Base`'s generic
  `AbstractArray` indexing rather than erroring, which is the intent.
- **The witness closes.** `𝕁 * StiefelManifold(...)` was a `MethodError` between two exported names
  and now returns `Matrix(𝕁) * Y.A`, as a `Matrix{Float64}`.
- **The `Strided…` bound really does cover the GPU arrays.**
  `MtlArray{T,N,S} <: AbstractGPUArray{T,N}` (`Metal/pSzWS/src/array.jl:46`) and
  `abstract type AbstractGPUArray{T, N} <: DenseArray{T, N}` (`GPUArraysCore/aNaXo/src/GPUArraysCore.jl:19`),
  so the `*` fast path is structurally safe on device. (I did not re-run the Metal measurement —
  see *Not checked*.)
- **The type-parameter bound bites.** `PoissonTensor{Float64, Matrix{Float32}}` is now a
  `TypeError`; `PoissonTensor{Float64}` still works.
- **The non-strided 3-tensor is a `MethodError`, as stated**, both through `*` and through the
  `QPTOAT` functor — and no call site in `src/`, `test/`, `scripts/` builds one.
- **No export moved.** `names(GeometricMachineLearning)` is byte-identical between `origin/main`
  and this branch: 126 names, `diff` empty. Dropping `import AbstractNeuralNetworks: add!` and
  `AbstractTriangular` changes no public surface.
- **Imports stay clean.** `improper_explicit_imports`: `STALE = 0`, `NOT-VIA-OWNER = 0`,
  `SELF-QUALIFIED = 0`, `NOT-VIA-OWNER qualified = 2 -> Gradient, direction` — matching the three
  gated checks and the `all_qualified_accesses_via_owners 2` note.
- **Every `file:line` citation into a changed file is correct.** The four that survive —
  `CHANGELOG.md:53,54,55` (`utils.jl:33`, `:39`, `:104`, read against `origin/main`'s `src/utils.jl`)
  and `CHANGELOG.md:388,853` (`src/GeometricMachineLearning.jl:83` = the `params, model, dim`
  import; `:179` = the `SymplecticStiefelLayer` note) — all land where they claim. The old
  `poisson_tensor.jl:42` and `gml_extensions.jl:17,22,27,32,39` citations went with their entries.
- **The three `[Unreleased]` passages that this branch contradicts are reconciled in place**, at
  `CHANGELOG.md:538`, `:1353` and `:1937`, rather than rewritten — which is right, since
  `[Unreleased]` is read as one release. `B8` is gone from `## Open Issues`; `B7` and `B9` remain.
- **Comment tense.** Every new or rewritten passage in the three source files and `test/aqua.jl` is
  in the present tense and describes the package as it is. Finding 3 is the only historical note
  left in a changed file, and it is outside the hunks.
- **Formatting and encoding.** JuliaFormatter `overwrite = false` over the five changed `.jl` files
  under the repo's `style = "sciml"`: all pass. `nfc.jl` over those five plus `CHANGELOG.md`:
  exit 0.

### CI — per job, not per workflow

The **workflow** conclusion for run `35434934045` is `success`. That is wrong, and it is the
`continue-on-error` case:

| job | status | conclusion |
|:--|:--|:--|
| Julia 1 — ubuntu / macOS / windows | completed | success |
| Julia min — ubuntu / macOS / windows | completed | success |
| Julia pre — ubuntu | completed | success |
| **Julia nightly — ubuntu** | completed | **failure** |
| Doctests — ubuntu | completed | success |
| Documentation | completed | success |
| Scripts — ubuntu | completed | success |
| codecov/patch, codecov/project | — | success |

The nightly log gives **one** failure and it is *B9*, not this branch:

```
Unbound type parameters detected:
Unbound type parameters: Test Failed at .../Aqua/yUvBM/src/unbound_args.jl:38
  Expression: isempty(unbounds)
...
Aqua's package-level checks | 13 pass 1 fail 14
  Aqua                      | 10 pass 1 fail 11
    Unbound type parameters |        1 fail  1
ERROR: LoadError: Some tests did not pass: 13 passed, 1 failed, 0 errored, 0 broken.
```

Both new gates ran and passed **on nightly too** — they sit after `unbound_args` inside the parent
`@testset`, and the surrounding 13 assertions are green. The `min` jobs passing is the useful part
for the `== 1` ambiguity gate: it holds on the 1.10 floor as well as on 1.13.

### Not checked

- **The full `Pkg.test()` figure** (76 testsets / 4360 passing / 2 broken). I did not re-run the
  suite locally; eight green CI jobs across three platforms and four Julia versions cover
  pass/fail, but not that exact count.
- **The Metal measurement.** I verified the *type* claim that makes it work
  (`MtlArray <: DenseArray <: StridedArray`) from the depot sources, but I did not run a device
  product. Finding 1 is about the path the Metal probe did **not** cover — the non-strided
  fall-through and ranged `getindex` — not about the strided one it did.
- **`#207`'s interaction.** The note in the PR body is correct that whichever lands second must
  re-derive the count from `Aqua.Piracy.hunt`. I did not review `#207`.

### Verdict

**Comment** — the substance is right and every claim I could re-measure reproduced exactly,
including all three "before" numbers. I would take findings 1 and 2 in this branch: 2 is a wrong
figure on a line the branch edited, and 1 is the one consequence of the narrowing that the PR's
otherwise careful accounting leaves out.

michakraus:
## Review findings resolved — `07118862`

Every finding from the [review above](https://github.com/JuliaGNI/GeometricMachineLearning.jl/pull/300#pullrequestreview-5255402139), by number. I re-measured each piece of quoted evidence before acting on it; all of it reproduced.

| # | severity | claim | action | evidence |
|--:|:---------|:------|:-------|:---------|
| 1 | correctness-risk | The narrowing moves both generic paths onto scalar `getindex`, which a GPU-backed `PoissonTensor` cannot serve | **fixed**, both halves | see below |
| 2 | quality | `test/aqua.jl` says `all_qualified_accesses_are_public` is 22; it is 24 | **fixed** | re-measured `improper_qualified_accesses(...)` counting `public_access === false`: **24**. The neighbouring `10` is correct and unchanged |
| 3 | quality | `src/utils.jl` opens with a historical note, which belongs in `CHANGELOG.md` | **fixed** | the two lines are removed |
| 4 | quality | *(mine, not in the review)* `## Open Issues` drops B8's entry without adding it to the "left vacant rather than reused" list at `CHANGELOG.md:3433` | **fixed** | B8 is now in that list |

### Finding 1, in detail

The reviewer's counting-wrapper numbers reproduced exactly — `𝕁[1:2, :]` 0 → 8 scalar reads, `𝕁[:, 1]` 0 → 4, `𝕁 * view(A, [1,2,3,4], :)` 0 → 48. I then checked it on a **real Apple GPU** rather than only a CPU stand-in, and it was a genuine regression: three expressions that answered on `main` raised `Scalar indexing is disallowed.`

Two changes close it.

**`getindex` names its index types** instead of being `::Int`-only. `_ForwardedIndex` is `Union{Int, Colon, AbstractVector{<:Integer}}`, and the whole index is forwarded to the wrapped array. Every type that caused one of the nine ambiguities is still excluded, and for a stated reason: `Block` and `BlockIndex` are neither integers nor vectors of them, and `BandRangeType` is a singleton marker.

**`ext/GPUArraysCoreExt.jl` restores the `*` path** for a *wrapped* GPU array — `view(A, [1,2,3,4], :)` and `A'` — which is not strided and so missed the `Strided…` methods. The bound is `AnyGPUArray`, not `AbstractGPUArray`: the wrapped case is precisely the one that failed. `GPUArraysCore` becomes a weak dependency at `"0.2"`, which is no new floor, since `AbstractNeuralNetworks` already pins it there.

On the device (`MetalBackend()`, `MtlArray`, `Float32`), every result equals the plain matrix product:

| expression | `main` | this branch before | now |
|:--|:--|:--|:--|
| `𝕁[1:2, :]` | OK | throws | **OK** |
| `𝕁[:, 1]` | OK | throws | **OK** |
| `𝕁[[1,3], [2,4]]` | OK | throws | **OK** |
| `𝕁 * view(A,[1,2,3,4],:)` | OK | throws | **OK** |
| `𝕁 * B'` | OK | throws | **OK** |
| `𝕁 * A` (strided) | OK | OK | OK |
| `𝕁 * (q,p)` | OK | OK | OK |

Scalar `𝕁[3, 1]` still raises on a device — it does on `main` too, so that is not a regression.

### Both gates still hold

Re-checked in a cold process with Zygote, GeometricIntegrators and HDF5 loaded, so BandedMatrices and BlockArrays are present — the configuration `test/aqua.jl` runs in — and with the extension active:

```
piracies (GeometricMachineLearning)    3
ambiguities (GeometricMachineLearning) 1   <- Dense/Affine, resnet.jl:63 vs affine.jl:11
ambiguities (GPUArraysCoreExt)         0
ambiguities (both together)            1
```

The extension module is scanned separately on purpose: `detect_ambiguities(GeometricMachineLearning)` alone does not reach it. For contrast, widening `*` back to `AbstractMatrix`/`AbstractVector` gives **17** in that same configuration, which is why the `Strided…` bound stays and the GPU path is restored by a narrower method set instead.

### What CI does and does not stand behind

**The GPU behaviour is not exercised by CI.** This repository has no GPU runner. It was verified on a local Apple GPU only.

The extension itself *does* load on CI, because `GPUArraysCore` arrives transitively through `AbstractNeuralNetworks`, so the new testset's two assertions — that the extension is loaded and defines its three methods — do run. The behaviour they gate is device-only, and the test comment says so. I did not use a CPU stand-in to imply device coverage.

The other new testset, "a bulk index reaches the wrapped array whole", is fully CI-exercisable: it uses a CPU wrapper that refuses scalar `getindex` the way the GPU libraries do. Each of its four assertions was checked individually against the pre-fix commit `158f26d9` — all four raise there, all four pass now.

### Not fixed

Nothing from the review is left open. Two notes:

- A non-strided **CPU** right-hand side still takes the generic multiply rather than the specialised one. Same value, no fast path. That is the intended cost of closing the seventeen ambiguities, and it is recorded under *Removed (breaking)*.
- A non-strided **CPU** 3-tensor is a `MethodError`. Nothing in the package, the tests or the scripts builds one.

PR #301, opened 2026-09-19
Documentation, comments and naming: audit Part F

michakraus:
**Part F** of the whole-package audit: documentation, comments and naming (§17–§19). Parts A–E are
merged as #292, #295, #297, #299 and #300.

## One breaking change

`SymplecticAttentionQ(M)` and `SymplecticAttentionP(M)` now default to `symmetric = true`. They
defaulted to `false` while the docstring above each interpolated a constant saying `true` — the
documented default was not the one a caller got. The prose in those same docstrings has always said
symmetric is almost always the better choice.

Nothing in the tree changes behaviour: `SymplecticTransformer`, the only caller, already passed
`symmetric = true` explicitly. An external caller of `SymplecticAttentionQ(4)` now gets a
`SymmetricMatrix` weight and `parameterlength` 3 instead of a plain matrix and 4. The keyword is
also typed `::Bool`, so a non-Bool is a `MethodError` rather than silently taking the `:arbitrary`
branch.

## The rest

- **Both `SymplecticAttention` constructors read their defaults off the constants**, so the
  docstring interpolation is true by construction. `SymplecticTransformer`'s docstring said
  `symmetric::Bool=false` where the constructor takes `true`, and that bullet had no description;
  `init_upper` and `symmetric` now interpolate their constants like the other four, and
  `n_sympnet` loses a stray paren that rendered its default as `(2)`.
- **Fourteen historical comment blocks** are rewritten or removed. Everything but `CHANGELOG.md`
  describes the package as it is. Where a block carried only the deletion it is gone; where it
  carried a present-tense fact, the fact stays and the history goes. **One of them was false** —
  `GeometricMachineLearning.jl` claimed "ten exported names are still undefined, which is issue
  C10", and C10 is closed by this release with an empty allowlist in `test/exports.jl`.
- **`symplectic_attention.jl`'s "for some reason we have to allocate a local variable here. This
  should be further investigated." is answered, not carried.** `A = ps.A` is load-bearing, and not
  for allocation: reading `ps.A` twice through the wrapper degrades the Zygote gradient leaf from a
  `SymmetricMatrix` to a plain `Matrix`. Measured — removing the binding makes
  `test/parameters/symplectic_attention_network_parameters_gradient.jl` fail with
  `Matrix{Float64} <: SymmetricMatrix` evaluated false. The new comment says that and points there.
- **One comment, copied into five kernel files, named the wrong operation in four of them.** Each
  now states its own operation as an indexed assignment.
- **Commented-out code** removed from six files, and five commented-out docstrings resolved.
  `assign_q_and_p.jl` and `sympnets.jl` were not on the audit's list and are the same defect found
  in the same sweep.
- **Seven `const` defaults** take the lowercase `<prefix>_<name>_default` spelling that the other
  31 in `src/` use. `hamiltonian_neural_network.jl`'s three are deliberately left: internally
  consistent, and #207 rewrites that file.
- **Typos**: `create_layers_for_transformer_dimension_eqaul_system_dimension` and its `uneqaul`
  sibling; the two softmax docstrings summed a term that does not depend on the summation index;
  `batch.jl`'s "this is neeced".
- **`TODO.md` is deleted.** Two of its three items are done — `LagrangianNeuralNetwork` is exported
  and tested, `PoissonTensor` generalises the symplectic matrix to any even dimension — and the
  third is recorded as C20.
- **A missing entry for an earlier change in this release.** Commit `914c72db` removed the piracy
  `AbstractNeuralNetworks.NetworkLoss(nn::NeuralNetwork)` on 2026-08-16. It had no changelog entry,
  only a source comment, and that comment goes here. It now has one.
- **Three new open issues**: C18 (the file and directory named after a `tensor_exponential` that
  neither holds), C19 (the two `_make_block_for_initialization` methods that differ only in which
  fields they read — this was a source `TODO`), C20.

## The new test, and its honest baseline

`test/attention/documented_defaults.jl` reads each keyword default off a constructed object and
compares it with the constant, for both attention layers and both transformer architectures.
Sixteen assertions. **Two fail on `main`** — the `symmetric` type parameter of each layer. The other
fourteen pass there: they are drift guards, not reproductions. A default written as a literal and
documented through a constant can be caught no other way.

## Pre-PR verification

Run by `julia-branch-verifier` against `git diff --name-status origin/main...fix/audit-part-f`.

**Fixed after the first pass**, all comments this branch wrote or touched:
- The two `SymplecticAttention` docstrings advised *setting* `symmetric` to `true`, which this
  change makes the default. They now say to leave it there.
- The `update!` note said GML imports `GeometricOptimizers.update!` "below"; the import is above it.
- "Three methods and not four" sat between two three-method families and named neither. It names
  `_euclidean_update!`.
- Two `file:line` citations in `CHANGELOG.md` into `src/GeometricMachineLearning.jl` that this
  branch falsified by deleting comment blocks above them: `:83` → `:80`, `:179` → `:169`.
- The `NetworkLoss` entry said three methods remain; there are four.
  `NetworkLoss(::AutoEncoder)` takes an unnamed argument, so the grep that found the other three
  missed it.
- The changelog attributed "three methods and not four" to `_go_update_leaf!`.
- Pre-existing but made contradictory by this branch: the `SymplecticAttention` docstring called
  the non-symmetric weight "skew-symmetric". It is an arbitrary matrix, as the
  `SymplecticTransformer` bullet rewritten here says.

**Checked and clean**
- Documenter HTML build reaches `RenderDocument` and `HTMLWriter` and exits 0, with no
  `missing docstring` and no `no docs found` — the docstring-detachment gate. `CheckDocument`
  completed. `docs/check_references.jl`: 0 unresolved references.
- Doctests pass. No doctest renders a keyword default.
- `Pkg.test()` on Julia 1.13 at `--check-bounds=auto`: **67 testsets, 4138 assertions, 0 failures,
  2 expected `broken`**, against the 64/4122/0/2 baseline plus this branch's 3 testsets and 16
  assertions.
- Aqua: `Piracy.hunt` = 3 and `detect_ambiguities` = 1, exactly what `test/aqua.jl` asserts.
  Unchanged — this branch adds no method on a foreign type.
- JuliaFormatter against the repository's own `style = "sciml"`: all changed files pass.
- No reference to any of the seven old constant names, to `eqaul`/`uneqaul`, or to `TODO.md`
  survives outside the changelog entries that record them — including no
  `GeometricMachineLearning.<name>` reference from outside `src/`, which is how Part E broke three
  call sites.
- Inference: the three affected constructors infer concretely, and `parameterlength` gains `Int64`
  where two abstract `::Integer` return annotations were removed.
- Zero history markers in any added source or test line.

**Pre-existing, reported and not fixed**
- `symmetric == false` at `src/layers/symplectic_attention.jl:78` and `:85` is `!symmetric` now
  that the keyword is typed. `fatou lint` `redundant-boolean`, advisory, identical on `main`.
- The `StandardTransformerIntegrator` field is spelled `upsacling_activation` while its keyword is
  `upscaling_activation`. Renaming the field is a breaking API change, so not a Part F edit — the
  new test has to spell the typo.
- Three file-top bare string literals are separated from their definition by a comment, so they are
  discarded expressions rather than docstrings: `tensor_tensor_transpose_mul.jl`,
  `tensor_transpose_tensor_mul.jl`, `tensor_assign.jl`. None of the three bindings is in a `@docs`
  block, so the docs build does not see it.
- The `compute_output_of_mha` comment heads the 3-tensor method but describes the function, as the
  commented-out docstring it replaces did.
- A docstring typo, `Zygote`-"induces" garbage, in `src/pullbacks/zygote_pullback.jl`.

**Not checked**
- The LaTeX manual and the PDF pipeline; only the HTML output was built.


michakraus (commented):
## Review — PR #301, audit Part F

Reviewed at head `d5035805111e620e83943203454d5620812c4501`, from a detached worktree against
`git diff --name-status origin/main...origin/pr/301` (26 files, not the truncating `--json files`
path).

**Verdict: approve.** No blocker and no bug. Every substantive claim I re-checked against the
source holds: the five kernel comments match their bodies index for index, the two softmax
summation indices are now correct, the seven renamed constants have no surviving reference outside
the changelog entries that record them, the `eqaul`/`uneqaul` rename leaves no stale caller, and
`TODO.md` is referenced nowhere but the changelog. The new test file is reached from the suite. The
findings below are quality and nit level: the sweep is narrower than the changelog reads, and four
comments of exactly the class this PR removes survive — three of them in the file the PR edited to
remove that class.

### Findings

| 1 | quality | `src/GeometricMachineLearning.jl:155-158`, `:169-172`, `:228-229`; `src/map_to_cpu.jl:5` | Four historical comments survive, three of them in the very file this PR edits to remove tombstones. The stated invariant — everything but `CHANGELOG.md` describes the package as it is — is not met. | `:155` "`UnknownProblem` and `NothingFunction` **were exported here**, under the comment 'are these needed?' … Both types **are gone** with the export." `:169` "`SymplecticStiefelLayer`, `ResidualLayer`, … **were exported here and defined nowhere**" — the same shape as the `SymplecticLieAlgMatrix` block this PR deletes at old `:123-128`. `:228` "GML's own `AdamOptimizerWithDecay` **was** a second, incompatible export of the same name — issue B1" — the same B1 history this PR strips from `src/optimizers/optimizer.jl`. `src/map_to_cpu.jl:5` "…which is precisely what the five per-type methods **this replaces were doing** by hand." |
| 2 | quality | `src/data_loader/tensor_assign.jl:55-57` | A commented-out docstring survives in a file this PR edits; eight such blocks remain in `src/`. | `grep -rn '^\s*#\s*"""' src/` returns 16 lines — eight paired blocks: `src/layers/stiefel_layer.jl:1,3`; `src/architectures/autoencoder.jl:128,130 / 145,147 / 152,154 / 169,173 / 180,184`; `src/architectures/symplectic_autoencoder.jl:117,119`; `src/data_loader/tensor_assign.jl:55,57`. `grep -rn '^\s*#\s*@doc' src/` returns nothing, so the changelog's scoping to `# @doc raw"""` blocks is literally accurate — but the bullet heads "Commented-out code is removed" and reads as closing the class. |
| 3 | quality | `src/data_loader/tensor_assign.jl:2` | A historical note survives two lines above a comment this PR rewrote. | Line 2: "A few data loader/tensor assignment functions **that were developed during Cemracs**." The PR does flag this file's top string as a discarded bare literal; it does not flag the history inside it. `grep -rni cemracs src/ docs/src/ test/` → this one line. |
| 4 | quality | `src/architectures/lagrangian_neural_network.jl:26-28` | The file now ends with a free-floating comment that explains what is *not* in it. | `wc -l` on the file is 28; lines 26-28 are the comment and nothing follows. The prose is present tense, but the block exists only because three definitions were removed above it, and it attaches to no definition in the file. A reader arriving at the file has no `∇L` to relate it to. |
| 5 | nit | `src/layers/symplectic_attention.jl:132-133` | The new comment names a wrapper that is not in scope at that method, and contradicts the test it points at. | The comment: "Read through **the wrapper** twice, the gradient leaf degrades to a plain `Matrix`". But `test/parameters/symplectic_attention_network_parameters_gradient.jl:45-50` — the test this comment cites — says the opposite about this route: "`Chain` reaches the layer through `values(_ps)` **rather than through a `getproperty` on the wrapper**, and on that route the count that decides the split is the one on **the plain `NamedTuple`** the layer is handed." Inside `(d::SymplecticAttentionQ{M,M,:symmetric})(z, ps)` the `ps` at `:138` is that plain `NamedTuple`. "twice on `ps`" says it without sending the reader after a wrapper. |
| 6 | nit | `src/layers/symplectic_attention.jl:78`, `:85` | `symmetric == false ? … : …` is redundant, and this PR is what made it so. | The PR types the keyword `::Bool` at `:74` and `:81`. Before that, `symmetric == false` was doing real work for a non-`Bool` argument. It is reported in the PR body as pre-existing and advisory; the typing change is what converts it into dead comparison, so it is diff-attributable now. `!symmetric` is the same three branches. |
| 7 | nit | `CHANGELOG.md` (Infrastructure, "Fourteen historical comment blocks") | The count is not reproducible from the diff. | Counting hunks in `git diff origin/main...origin/pr/301` whose removed lines narrate history, I get 18: `src/GeometricMachineLearning.jl` ×6 (`@@31`, `@@64`, `@@73`, `@@123`, `@@214`, `@@231`), `optimizers/optimizer.jl` ×5 (the alias, `AdamOptimizerWithDecay`, `_step_size`, `_euclidean_update!`, `optimization_step!`), and one each in `lagrangian_neural_network.jl`, `tensor_assign.jl`, `tensor_exponential.jl`, `sympnets.jl`, `losses.jl` ×2, `zygote_pullback.jl`. Either the number or the grouping rule needs stating. |
| 8 | nit | `CHANGELOG.md` (Removed (breaking), `SymplecticAttention` default) | "The only in-tree caller, `SymplecticTransformer`, already passed `symmetric = true` explicitly" — it does not pass a literal. | `src/architectures/symplectic_transformer.jl:90` and `:94` pass `symmetric = arch.symmetric`. The value is `true` only because `st_symmetric_default` at `:6` is `true`. The conclusion (no behaviour change in the tree) is right; the stated reason is one step short. Confirmed no other call site: `grep -rn 'SymplecticAttentionQ(\|SymplecticAttentionP(' src/ test/ docs/ scripts/` finds those two plus `test/parameters/symplectic_attention_network_parameters_gradient.jl:52`, which passes `symmetric = true` explicitly. |
| 9 | nit | `CHANGELOG.md` (section placement) | A default-value flip is filed under `### Removed (breaking)`; nothing is removed. | `grep -n '^### ' CHANGELOG.md`: the Unreleased block has `Removed (breaking)` at `:41` and `Changed` at `:592`, no `Changed (breaking)`. The 0.5.0 block has `Changed (breaking)` at `:3136`, so the heading exists in this file's vocabulary. The Unreleased `Removed (breaking)` section already holds a rename (`_add(::History, ::SingleHistory)` → `_push_history!`), so this is internally consistent — but a reader scanning headings for a behaviour change will not look there. |
| 10 | nit | `src/architectures/standard_transformer_integrator.jl:21-22`, `:29-30` | Two bullets render a stray space before their colon — the same class as the `n_sympnet` stray paren this PR fixes, on a line this PR touched. | `:29` is `"$(sti_attention_activation_default)` "` — a space inside the interpolated fragment after the closing backtick — so the bullet renders as ``` `attention_activation = VectorSoftmax()` : the activation … ``` . `:22` has the same space at the start of its `raw"""` fragment. Pre-existing; `:29` is rewritten by this diff. |
| 11 | nit | `test/attention/documented_defaults.jl` | The test closes one half of the drift it describes, not both. | The file compares *constructed object* against *constant*. The docstring-to-constant half is unguarded: replacing `$(sa_symmetric_default)` in `src/layers/symplectic_attention.jl:46` with the literal `true` leaves all 16 assertions green. That is fine — nothing cheap tests it — but the changelog's "Only a test that constructs the object and reads the default it actually got can catch that" claims the whole loop. |

### Verified and fine

- **The new test is reached.** `test/runtests.jl:33` includes `attention/runtests.jl`; `test/attention/runtests.jl:30-32` adds the `@safetestset` that includes `documented_defaults.jl`. Sixteen assertions, matching the claim: 2 constructors × 2, plus 6 and 6.
- **Every field and constant the new test names exists, with the type it assumes.** `SymplecticTransformer` carries `n_sympnet`, `L`, `sympnet_activation`, `attention_activation`, `init_upper`, `symmetric` as fields; `SymplecticAttention` carries `symmetric` as its fourth type parameter (`:symmetric` / `:arbitrary`), which is what `typeof(l).parameters[4]` reads. `StandardTransformerIntegrator`'s field really is spelled `upsacling_activation`, so the test spelling is right.
- **The five kernel comments are correct, index for index.** `mat_tensor_mul` `A[i,l]*B[l,j,k]`; `tensor_mat_mul` `A[i,l,k]*B[l,j]`; `tensor_tensor_mul` `a[i,l,k]*b[l,j,k]`; `tensor_tensor_transpose_mul` `A[i,l,k]*B[j,l,k]` = `(A_k B_k')_{ij}`; `tensor_transpose_tensor_mul` `A[l,i,k]*B[l,j,k]` = `(A_k' B_k)_{ij}`.
- **The deleted C10 comment really was false.** `test/exports.jl:13` is `const UNDEFINED_EXPORTS = Dict{Symbol, String}()` — empty — and `:24` asserts the allowlist may not outlive what it excuses.
- **`assign_q_and_p` returns a `NamedTuple`, as the new comment now says.** `src/kernels/assign_q_and_p.jl:48`, `:59` return `(q = q, p = p)`; the commented-out docstring it replaces said "a `Tuple`".
- **No stale reference survives.** `grep -rn` over the tree for `TODO.md`, for `eqaul`/`uneqaul`, and for the seven old constant spellings returns hits in `CHANGELOG.md` only.
- **No new docstring detachment.** Every comment this diff adds above a definition follows an `end` or a file start, not a docstring. The three pre-existing file-top bare strings are unchanged, and the `Documentation` job passed.
- **Every `file:line` citation the diff adds or updates in `CHANGELOG.md` resolves at head**: `src/GeometricMachineLearning.jl:80` (the `dim` import), `:169` (the undefined-exports comment), `src/loss/hnn_loss.jl:29`, `src/loss/lnn_loss.jl:55` and `:122`, `src/loss/losses.jl:59` and `:146`.
- **`SymplecticAttention` is in a `@docs` block** (`docs/src/layers/symplectic_attention.md:88-93`), so the `@ref` the new `symmetric` bullet adds resolves. No page in `docs/src/` states the old `false` default.
- **JuliaFormatter against the repository's own `style = "sciml"`**: all 24 changed `.jl` files are already formatted (`format_text` returns the file unchanged for each).
- **No closing keyword in the PR body** — `gh pr view --json closingIssuesReferences` is `[]`, so nothing auto-closes on merge.

### CI, per job at `d503580`

| job | conclusion | reading |
|:--|:--|:--|
| Julia 1 — ubuntu / macOS / windows | success | |
| Julia min — ubuntu / macOS / windows | success | |
| Julia pre — ubuntu | success | |
| Julia nightly — ubuntu | **failure** | **Not this PR.** Aqua "Unbound type parameters", 13 passed / 1 failed, at `Aqua/src/unbound_args.jl:38`. The nightly job on the `main` tip fails on the same assertion with the same 13/1 split. Pre-existing on 1.14-DEV. |
| Doctests — ubuntu | success | |
| Documentation | success | 59m, completed. This is the docstring-detachment gate. |
| Scripts — ubuntu | success | |
| codecov/project | success | 85.35%, +0.23% against the base. |
| codecov/patch | **failure** | **An artefact of the rename, not a coverage regression.** 42.85% of diff hit; the 4 missing lines are all in `src/architectures/symplectic_transformer.jl` — the four renamed `const` lines at the top of the file, which coverage never instruments because they run during precompilation. Total line count is unchanged at 3012 and hits go **up** by 7. |

### Not checked

- **The PR body's inference claim** — "the three affected constructors infer concretely". I did not measure it, and I would want the measurement restated before trusting it: `SymplecticAttentionQ(M::Integer)` puts the runtime *value* `M` into a type parameter and annotates `activation::AbstractSoftmax`, so a concrete result depends on constant propagation from a literal argument. `@inferred` drops literal arguments, so it cannot establish this without a closure. Not a finding — the claim is in the PR body, not in the diff.
- **The `A = ps.A` mechanism was not re-measured.** The diff does not touch `test/parameters/symplectic_attention_network_parameters_gradient.jl`; the test and its explanatory comment are pre-existing, and the three testsets there do assert both halves (two accesses → structure lost, one access → kept) plus the layer itself. Finding 5 is about the wording of the new source comment, not about the mechanism.
- **The LaTeX manual and the PDF pipeline**, as the PR says. Only the HTML `Documentation` job ran.
- **`docs/check_references.jl`** was not run locally; I relied on the green `Documentation` job and on reading the one `@ref` target the diff adds.

michakraus:
## Review findings resolved, in `5f4288e8` and `97b6c712`

Every finding from the review above, plus two further pieces of work requested afterwards.

### The eleven findings

| # | severity | where | action | evidence |
|--:|:---------|:------|:-------|:---------|
| 1 | quality | `src/GeometricMachineLearning.jl`, `src/map_to_cpu.jl` | fixed | Two tombstone blocks deleted; the `AdamOptimizer` re-export comment keeps its present-tense sentence and loses the B1 history; the `map_to_cpu` clause "which is precisely what the five per-type methods this replaces were doing by hand" is gone. Export count unchanged at 76 on both sides, and the diff for that file is comments only. |
| 2 | quality | `src/data_loader/tensor_assign.jl` | fixed | The commented-out docstring above `augment_zeros_kernel!` becomes a plain present-tense comment, matching what this branch already did for `compute_output_of_mha` and `apply_layer_to_nt_and_return_array`. The other seven blocks of this class were swept separately — see below. |
| 3 | quality | `src/data_loader/tensor_assign.jl` | fixed | "that were developed during Cemracs" removed from the file-top string. |
| 4 | quality | `src/architectures/lagrangian_neural_network.jl` | fixed | The trailing free-floating block is deleted. The fact it carried is already stated at `src/loss/lnn_loss.jl:35-37`, so nothing is lost. |
| 5 | nit | `src/layers/symplectic_attention.jl` | fixed | The comment said "Read through the wrapper twice". At that method `ps` is a plain `NamedTuple`; `test/parameters/symplectic_attention_network_parameters_gradient.jl` says `Chain` reaches the layer through `values(_ps)` rather than a `getproperty` on the wrapper. Now reads "Read `ps.A` twice". |
| 6 | nit | `src/layers/symplectic_attention.jl` | fixed | `symmetric == false` becomes `!symmetric` in both constructors. `fatou lint` on the file is clean, exit 0, and the `redundant-boolean` it cited is gone. |
| 7 | nit | `CHANGELOG.md` | fixed | Re-measured rather than taken on trust. Counting maximal runs of removed line-initial `#` comment lines carrying past-tense history in `git diff main...HEAD -- src/` gives **17** — 6 in `src/GeometricMachineLearning.jl`, 5 in `src/optimizers/optimizer.jl`, 6 elsewhere. The review's hunk-based count gave 18. Neither reproduces 14, and since two careful counts disagree the entry no longer asserts a number. |
| 8 | nit | `CHANGELOG.md` | fixed | Corrected: `SymplecticTransformer` passes `symmetric = arch.symmetric` at `src/architectures/symplectic_transformer.jl:90` and `:94`, and that value is `true` only through `st_symmetric_default` at `:6`. |
| 9 | nit | `CHANGELOG.md` | fixed | A new `### Changed (breaking)` heading carries the default flip, placed after `### Changed` to match the ordering the `[0.5.0]` block uses. |
| 10 | nit | `src/architectures/standard_transformer_integrator.jl` | partly fixed | The `attention_activation` bullet's stray space is removed. The sibling defect on the `n_blocks` bullet — `raw""" : ` — is left: it sits outside every hunk this branch touches. |
| 11 | nit | `test/attention/documented_defaults.jl` | not fixed | Confirmed by running it, not by inspection. With the `SymplecticAttentionQ` docstring's interpolated constant replaced by a literal `false` — a docstring then actively contradicting the real default — all 16 assertions still passed. The gap is real. It is left because closing it needs a design decision rather than an edit: the obvious form, asserting on the rendered docstring, is defeated here by the prose in that same docstring, which contains the word `true` independently of the interpolation. |

### Two further changes, requested after the review

Neither is diff-attributable; both were asked for explicitly rather than being defects this branch
introduced. The same is true of findings 1, 2 and 3. Findings 4 and 5–11 are this branch's own —
finding 4 in particular, whose block was written by this branch in hunk `@@ -23,9 +23,6 @@`
replacing an older historical comment, so the review's reading of it as pre-existing is not right.

**A test for `Chain(::SymplecticTransformer)`.** Neither method had one, and with them neither
`create_layers_for_transformer_dimension_equal_system_dimension` nor its `unequal` sibling.
`test/transformers/symplectic_transformer_chain.jl` builds both shapes and asserts the layer count
and the sequence of layer types, and is wired into `test/transformers/runtests.jl`. The expected
counts are read off the two builders rather than off a run — with the defaults `L = 1` and
`n_sympnet = 2` the equal-dimension builder emits `L * (2 + 2 * n_sympnet) = 6`, and the unequal one
wraps that in a `PSDLayer` at each end for 8. Predicted before running; the run agreed. 12
assertions.

**The remaining seven plain `# """` blocks are swept.** `grep -rn '^\s*#\s*"""' src/` now returns
nothing. The split is not what was expected, so it is worth stating:

- **Deleted (3).** `src/layers/stiefel_layer.jl`, whose `x ↦ Ax` behaviour `ManifoldLayer` already
  documents on the parent type; and the two in `src/architectures/autoencoder.jl` on
  `encoder_parameters` and `decoder_parameters`. Those last two are the blocks carrying a signature
  line, which might look like the ones most worth keeping — but each one's signature line is
  repeated verbatim by the `function` line directly below it, and its prose restates the name, so
  they are the most deletable of the seven.
- **Converted to ordinary comments (4)**, each carrying a fact the signature does not:
  `compute_iterations` (ascending order); `encoder_layers_from_iteration` and
  `decoder_layers_from_iteration` (an implementation must return a tuple of
  `AbstractExplicitLayer`s — both are `error(...)` interface stubs, so that contract has nowhere
  else to live); and `compute_iterations_for_symplectic_system` (ascending order, even steps only).

### Left alone, deliberately

The `n_blocks` bullet's `raw""" : ` in `src/architectures/standard_transformer_integrator.jl`. It is
the same class as finding 10 but outside every hunk this branch touches.

### Verification

- The per-assertion baseline for `test/attention/documented_defaults.jl` was re-measured on `main`
  rather than taken from the description: 16 assertions, exactly 2 failing, and the two are
  precisely the `symmetric` type-parameter assertions for `SymplecticAttentionQ` and
  `SymplecticAttentionP`. The other fourteen pass there, so they are drift guards as described.
- After every edit: the package loads, all 16 of those assertions pass, JuliaFormatter reports the
  changed `.jl` files clean against the repository's own `sciml` style, and the NFC check exits 0.
- Each deleted comment block was checked for docstring attachment: every one sat after an `end` and
  a blank line, so none of the deletions moved a real docstring onto a new definition. That is a
  local check; the `Documentation` job is what can actually confirm it.
- `fatou lint` reports 25 `unused-import` warnings in `src/GeometricMachineLearning.jl`. The count
  is identical on `main`, and the rule does not follow `include`, so they are pre-existing false
  positives rather than anything these changes introduced.

### CI at `97b6c712`

All seven required checks pass — `Doctests - ubuntu-latest`, and `Julia 1` and `Julia min` on
ubuntu, macOS and windows. `Documentation` is green, which is the only check that can catch a
docstring detached by the comment deletions above. `Scripts`, `Julia pre` and `codecov/project`
are green, and `codecov/patch` has gone from failing to passing.

`Julia nightly - ubuntu-latest` fails, and is not required. It fails on Aqua's
`Unbound type parameters` assertion, 1 of 1, which fails the same way on the base branch — it is
not caused by anything here.

